PR #18, opened 2026-02-14
nv-redfish: Support WIWYNN non-standard response of chassis assembly

poroh:
Support of WIWYNN's non-standard response on Assembly resource in Chassis which doesn't contain required "@data.type" field.

### DIFF
--- a/redfish/src/patch_support/payload.rs
+++ b/redfish/src/patch_support/payload.rs
@@ -32,6 +32,9 @@ use nv_redfish_core::Updatable;
 #[cfg(feature = "patch-payload-update")]
 use serde::Serialize;
 
+#[cfg(feature = "patch-payload-expand")]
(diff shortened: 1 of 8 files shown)


redfish/src/patch_support/payload.rs
```diff
@@ -84,6 +86,28 @@ impl Payload {
             }
         }
     }
+
+    #[allow(dead_code)] // enabled by features
```
yoks: Is there any other way to make it work without linter allow?  If it enabled by features maybe annotate with set of features which enable it? Also make it a little bit clearer to see what features involved

poroh: Added feature-gating for patch functions as separate PR:  #19 .  This one in draft (will be merged on top of 19).

PR #67, opened 2026-04-03
csdl-compiler: redfish: fallback enum variant for unsupported values

poroh:
Generate `UnsupportedValue` for Redfish enums and use it as a serde fallback for unknown schema values. This makes enum deserialization more forward-compatible and removes the need for several ad hoc payload patches that previously sanitized invalid enum members.

Update SSE EventType handling to patch only missing values and map them to `UnsupportedValue` instead of coercing unknown values to `Other`.

Also clean up now-unneeded enum-sanitizing quirks/patches and tighten feature gating around payload patch support.

Tests cover unknown enum fallback and existing vendor cases that return unsupported enum values.


poroh:
Notes:
1. This PR adds conformance with "9.5.2 Enumerations" part of Redfish specification: "Client software should be prepared to receive enumeration values that are not known if the resource schema version is higher than the client's supported version."
2. After this PR merged we need to bump minor version because we additional field is added in bunch of generated and published enums.

yoks (approved):

PR #75, opened 2026-04-19
core: make EdmDateTimeOffset to SystemTime conversion fallible

poroh:
SystemTime cannot represent arbitrary time points on all platforms. On Windows it is backed by FILETIME, whose epoch is 1601, so any earlier time point is not representable and the previous infallible conversion would panic.

Replace the `From<EdmDateTimeOffset> for SystemTime` impl with a `TryFrom` impl that uses `checked_add`/`checked_sub` and returns a dedicated `Error::OutOfSystemTimeRange` variant when the value falls outside the representable range.

Update tests accordingly; the year 0001 case is now only asserted on platforms where the conversion succeeds.

### DIFF
--- a/core/src/edm_date_time_offset.rs
+++ b/core/src/edm_date_time_offset.rs
@@ -49,6 +49,8 @@
 
 use core::str::FromStr;
 use serde::{Deserialize, Serialize};
+use std::convert::TryFrom;
+use std::error::Error as StdError;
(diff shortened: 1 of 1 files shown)


poroh:
This is breaking change. Infallible conversions of EdmDateTimeOffset to SystemTime is removed.

yoks (approved):

PR #97, opened 2026-05-16
feat(redfish): add UpdateService upload APIs

jayzhudev:
## Description

Adds Redfish UpdateService multipart upload support.

- Extends the BMC and HTTP client traits with multipart UpdateService upload support.
- Streams UpdateFile from any async reader through a named DataStream.
- Uses the generated UpdateParametersUpdate schema type through the MultipartUpdateParameters alias for the normal high-level API.
- Uses the service-provided MultipartHttpPushUri from the UpdateService resource.
- [Spec](https://www.dmtf.org/sites/default/files/standards/documents/DSP0266_1.23.0.html#multipart-http-push-updates) conformant parts including optional OEM parts support.

## Type of Change

- [x] **Add** - New feature or capability
- [ ] **Change** - Changes in existing functionality
- [ ] **Fix** - Bug fixes
- [ ] **Remove** - Removed features or deprecated functionality
- [ ] **Internal** - Internal changes (refactoring, tests, docs, etc.)

## Breaking Changes

- [x] This PR contains breaking changes

Adds required methods to `Bmc` and `HttpClient`. Implementations must implement multipart update support.

## Testing

- [x] Unit tests added/updated
- [x] Integration tests added/updated
- [ ] Manual testing performed
- [ ] No testing required (docs, internal refactor, etc.)

## Additional Notes

Multipart upload is reader/stream-based so callers can provide file-backed streams or other data sources.

Signed-off-by: Jay Zhu <jayzhu@nvidia.com>

### DIFF
--- a/bmc-http/src/lib.rs
+++ b/bmc-http/src/lib.rs
@@ -45,7 +45,14 @@ pub mod credentials;
 #[cfg(feature = "reqwest")]
 pub mod reqwest;
 
+use std::collections::HashMap;
+use std::error::Error as StdError;
+use std::future::Future;
+use std::sync::Arc;
+use std::sync::RwLock;
+
 use crate::cache::TypeErasedCarCache;
+
 use http::HeaderMap;
 use nv_redfish_core::query::ExpandQuery;
 use nv_redfish_core::Action;
@@ -58,16 +65,15 @@ use nv_redfish_core::ModificationResponse;
 use nv_redfish_core::ODataETag;
 use nv_redfish_core::ODataId;
 use nv_redfish_core::SessionCreateResponse;
+use nv_redfish_core::UploadReader;
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
-use std::collections::HashMap;
-use std::error::Error as StdError;
-use std::future::Future;
-use std::sync::Arc;
-use std::sync::RwLock;
 use url::Url;
 
 #[doc(inline)]
 pub use credentials::BmcCredentials;
+
+#[doc(inline)]
+pub use nv_redfish_core::MultipartUpdateRequest;
 
 /// HTTP Client trait.
 ///
@@ -112,6 +118,22 @@ pub trait HttpClient: Send + Sync {
         B: Serialize + Send + Sync,
         T: DeserializeOwned + Send + Sync;
 
+    /// Performs an UpdateService multipart upload with credentials and headers.
+    ///
+    /// The request carries `UpdateParameters`, `UpdateFile`, and optional OEM
+    /// multipart parts.
+    fn post_multipart_update<U, V, T>(
+        &self,
+        url: Url,
+        request: MultipartUpdateRequest<'_, U, V>,
+        credentials: &BmcCredentials,
+        custom_headers: &HeaderMap,
+    ) -> impl Future<Output = Result<ModificationResponse<T>, Self::Error>> + Send
+    where
+        U: UploadReader,
+        T: DeserializeOwned + Send + Sync,
+        V: Serialize + Send + Sync;
+
     /// Perform an HTTP PATCH request.
     fn patch<B, T>(
         &self,
@@ -153,7 +175,6 @@ pub trait HttpClient: Send + Sync {
 /// # Type Parameters
 ///
 /// * `C` - The HTTP client implementation to use
-///
 pub struct HttpBmc<C: HttpClient> {
     client: C,
     redfish_endpoint: RedfishEndpoint,
@@ -552,6 +573,31 @@ where
             .post(
                 endpoint_url,
                 params,
+                credentials.as_ref(),
+                &self.custom_headers,
+            )
+            .await
+    }
+
+    async fn multipart_update<U, V, R>(
+        &self,
+        uri: &str,
+        request: MultipartUpdateRequest<'_, U, V>,
+    ) -> Result<ModificationResponse<R>, Self::Error>
+    where
+        U: UploadReader,
+        R: Send + Sync + for<'de> Deserialize<'de>,
+        V: Send + Sync + Serialize,
+    {
+        // MultipartHttpPushUri can be absolute or BMC-relative.
+        // Match existing URI handling before adding auth and headers.
+        let endpoint_url = Url::parse(uri).unwrap_or_else(|_| self.redfish_endpoint.with_path(uri));
+        let credentials = self.read_credentials();
+
+        self.client
+            .post_multipart_update(
+                endpoint_url,
+                request,
                 credentials.as_ref(),
                 &self.custom_headers,
             )
--- a/bmc-http/src/reqwest.rs
+++ b/bmc-http/src/reqwest.rs
@@ -15,26 +15,36 @@
 
 //! Implementation of [`HttpClient`] trait using reqwest crate.
 
+use std::error::Error as StdErr;
+use std::fmt;
+use std::time::Duration;
+
 use crate::BmcCredentials;
 use crate::CacheableError;
 use crate::HttpClient;
+use crate::MultipartUpdateRequest;
+
 use futures_util::StreamExt as _;
-use http::header;
 use http::HeaderMap;
+use http::header;
 use nv_redfish_core::AsyncTask;
 use nv_redfish_core::BoxTryStream;
+use nv_redfish_core::DataStream;
 use nv_redfish_core::ModificationResponse;
 use nv_redfish_core::ODataETag;
 use nv_redfish_core::ODataId;
+use nv_redfish_core::OemMultipartPart;
 use nv_redfish_core::SessionCreateResponse;
-use reqwest::redirect::Policy as RedirectPolicy;
+use nv_redfish_core::UploadReader;
 use reqwest::Client as ReqwestClient;
 use reqwest::Error as ReqwestError;
-use serde::de::DeserializeOwned;
+use reqwest::multipart::Form;
+use reqwest::multipart::Part;
+use reqwest::redirect::Policy as RedirectPolicy;
 use serde::Serialize;
-use std::error::Error as StdError;
-use std::fmt;
-use std::time::Duration;
+use serde::de::DeserializeOwned;
+use tokio_util::compat::FuturesAsyncReadCompatExt as _;
+use tokio_util::io::ReaderStream;
 use url::Url;
 
 /// Errors of reqwest implementation of the HTTP trait.
@@ -61,6 +71,10 @@ pub enum BmcError {
     CacheError(String),
     /// JSON deserialization error.
     DecodeError(serde_json::Error),
+    /// JSON serialization error.
+    EncodeError(serde_json::Error),
+    /// Invalid request error - data in the request didn't pass validation.
+    InvalidRequest(String),
 }
 
 impl From<reqwest::Error> for BmcError {
@@ -107,17 +121,19 @@ impl fmt::Display for BmcError {
             ),
             Self::SseStreamError(e) => write!(f, "SSE stream decode error: {e}"),
             Self::DecodeError(e) => write!(f, "JSON Decode error: {e}"),
+            Self::EncodeError(e) => write!(f, "JSON Encode error: {e}"),
+            Self::InvalidRequest(e) => write!(f, "Invalid request: {e}"),
         }
     }
 }
 
-impl StdError for BmcError {
-    fn source(&self) -> Option<&(dyn StdError + 'static)> {
+impl StdErr for BmcError {
+    fn source(&self) -> Option<&(dyn StdErr + 'static)> {
         match self {
             Self::ReqwestError(e) => Some(e),
             Self::JsonError(e) => Some(e.inner()),
             Self::SseStreamError(e) => Some(e),
-            Self::DecodeError(e) => Some(e),
+            Self::DecodeError(e) | Self::EncodeError(e) => Some(e),
             _ => None,
         }
     }
@@ -264,7 +280,6 @@ impl ClientParams {
 /// This provides a concrete implementation of [`HttpClient`] using the popular
 /// reqwest HTTP client library. It supports all standard HTTP features including
 /// TLS, authentication, and connection pooling.
-///
 #[derive(Clone)]
 pub struct Client {
     client: ReqwestClient,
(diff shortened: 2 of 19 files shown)


redfish/src/update_service/mod.rs
```diff
@@ -30,21 +30,146 @@ use crate::NvBmc;
...
 #[doc(inline)]
 pub use software_inventory::VersionRef;
 
+/// Common Redfish UpdateService multipart upload URI.
+#[cfg(feature = "bmc-http")]
+pub const UPDATE_MULTIPART_URI: &str = "/redfish/v1/UpdateService/update-multipart";
+
+/// Common Redfish UpdateService raw file upload URI.
+#[cfg(feature = "bmc-http")]
+pub const UPDATE_URI: &str = "/redfish/v1/UpdateService/update";
```
poroh: Different platforms may define these URIs differently. Redfish provides properties with value of these properties:

UPDATE_MULTIPART_URI value must be taken from `MultipartHttpPushUri` property (See:
https://docs.rs/nv-redfish/latest/nv_redfish/schema/update_service/struct.UpdateService.html#structfield.multipart_http_push_uri )

UPDATE_URI value from `HttpPushUri` (https://docs.rs/nv-redfish/latest/nv_redfish/schema/update_service/struct.UpdateService.html#structfield.http_push_uri)

These values can be accessed using data property of UpdateService struct.

redfish/src/update_service/mod.rs
```diff
@@ -30,21 +30,146 @@ use crate::NvBmc;
...
+    /// Redfish target resource URIs to update.
+    #[serde(rename = "Targets")]
+    pub targets: Vec<String>,
+    /// Additional Redfish or vendor-specific update parameters.
+    ///
+    /// Use [`Self::with_parameter`] to add entries without colliding with typed
+    /// `ForceUpdate` or `Targets` fields.
+    #[serde(flatten)]
+    additional_parameters: serde_json::Map<String, JsonValue>,
+}
```
poroh: It seems that this is `UpdateParameters` complex type that is defined by schema here:
https://github.com/DMTF/Redfish-Publications/blob/26c949654e3e510800b05fa8d39cf336f10fa687/csdl/UpdateService_v1.xml#L1016

In nv-redfish we avoid manually implement data structures that are defined by standard. However, CSDL compiler today doesn't have way to bring this to you because it is defined as ComplexType and this type is not attached to the Redfish tree. We had similar approach to this for EntityType (root_patterns). I guess that we should extend it to complex types as well. I'll do it as separate PR.

After my PR of extending root_patterns you can avoid defining these parameters.

redfish/src/update_service/mod.rs
```diff
@@ -247,6 +372,131 @@ impl<B: Bmc> UpdateService<B> {
     }
 }
 
+#[cfg(feature = "bmc-http")]
+impl UpdateService<HttpBmc<ReqwestClient>> {
```
poroh: nv-redfish consistently implements everything through BMC trait. This approach opens possibility to write testable code without need to bring up mock HTTP server. So, we need to keep it this way.

I don't know full upload service business so far but if we need to extend BMC trait to accommodate requirements we need to do so.

poroh (changes requested):

redfish/src/update_service/mod.rs
poroh: #98 

Generated type:
```
    #[derive(Deserialize, Debug)]
    pub struct UpdateParameters {
        /// An array of URIs that indicate where to apply the update image.
        ///
        /// This property shall contain zero or more URIs that indicate where to apply the update image when
        /// using the URI specified by the `MultipartHttpPushUri` property to push a software image.  These
        /// targets should correspond to software inventory instances or their related items.  If this property
        /// is not present or contains no targets, the service shall apply the software image to all applicable
        /// targets, as determined by the service.  If the target specifies a device resource, the software
        /// image file shall be applied to the specified device.  If the target specifies a resource
        /// collection, the software image shall be applied to each applicable member of the specified
        /// collection.  If the target resource specifies an `Aggregate` resource, the software image file
        /// shall be applied to each applicable element of the specified aggregate.  If the target resource
        /// specifies a `ComputerSystem` resource, the software image file shall be applied to the applicable
        /// components within the specified computer system.
        #[serde(rename = "Targets", default, deserialize_with = "de_optional_nullable")]
        pub targets: Option<Option<Vec<redfish::edm::String>>>,
        /// The OEM extension property.
        ///
        /// This property shall contain the OEM extensions.  All values for properties contained in this object
        /// shall conform to the Redfish Specification-described requirements.
        #[serde(rename = "Oem", default)]
        pub oem: Option<redfish::resource::Oem>,
        /// An indication of whether the service should bypass update policies when applying the provided
        /// image.  The default is `false`.
        ///
        /// This property shall indicate whether the service should bypass update policies when applying the
        /// provided image, such as allowing a component to be downgraded.  Services may contain update
        /// policies that are never bypassed, such as minimum version enforcement.  If the client does not
        /// provide this parameter, the service shall default this value to `false`.
        #[serde(rename = "ForceUpdate", default)]
        pub force_update: Option<redfish::edm::Boolean>,
        /// An indication of whether the service stages the image on target devices for a client to activate at
        /// a later time.
        ///
        /// This property shall indicate whether the service stages the image on target devices for a client to
        /// activate at a later time with the `Activate` action on the `SoftwareInventory` resource.  If the
        /// client does not provide this parameter, the service shall default this value to `false`.
        #[serde(rename = "Stage", default)]
        pub stage: Option<redfish::edm::Boolean>,
        /// An indication of whether the service adds the image to the local image store.
        ///
        /// This property shall indicate whether the service adds the image to the resource collection
        /// referenced by the `LocalImageStore` property.  If the client does not provide this parameter, the
        /// service shall default this value to `false`.
        #[serde(rename = "LocalImage", default)]
        pub local_image: Option<redfish::edm::Boolean>,
        /// An array of URIs that indicate where not to apply the update image.
        ///
        /// This property shall contain zero or more URIs that indicate where not to apply the update image.
        /// This property shall be ignored if the `Targets` property is provided and contains at least one
        /// member.  These excluded targets should correspond to software inventory instances or their related
        /// items.  If this parameter is not present or contains no targets, the service shall apply the
        /// software image to all applicable targets, as determined by the service.  If an excluded target
        /// specifies a device resource, the software image file shall not be applied to that specified device.
        ///  If an excluded target specifies a resource collection, the software image shall not be applied to
        /// each applicable member of the specified collection.  If an excluded target resource specifies an
        /// `Aggregate` resource, the software image file shall not be applied to each applicable element of
        /// the specified aggregate.  If an excluded target resource specifies a `ComputerSystem` resource, the
        /// software image file shall not be applied to the applicable components within the specified computer
        /// system.
        #[serde(rename = "ExcludeTargets", default)]
        pub exclude_targets: Option<Vec<redfish::edm::String>>,
    }
```

jayzhudev: Great, thanks!

redfish/src/update_service/mod.rs
jayzhudev: Thanks for the comment! Multipart update is used for firmware uploads through Redfish. We can:

- extend the Bmc trait -> breaking change, or
- use Bmc trait as a supertrait for this -> testable without HTTP server and no breaking change

Which way would suite this crate better?

redfish/src/update_service/mod.rs
jayzhudev: I'll use the build-time generated code after your PR as reference to reduce manual definitions of structs.

jayzhudev: I've rebased this PR on #98 and used the generated struct. This PR should only be merged after #98 is merged.

redfish/src/update_service/mod.rs
jayzhudev: Added a trait UpdateServiceUploadBmc to extend Bmc supertrait. Code is testable without a mock HTTP server.

redfish/src/update_service/mod.rs
jayzhudev: Thanks for catching this. Agreed, using UpdateService's data properties now.

core/src/bmc.rs
```diff
@@ -65,6 +65,46 @@ use crate::SessionCreateResponse;
...
+
+use futures_io::AsyncRead;
+
+/// Async reader accepted by upload methods.
+pub trait UploadReader: AsyncRead + Send + 'static {}
+
+impl<T> UploadReader for T where T: AsyncRead + Send + 'static {}
+
+/// Named data stream accepted by upload methods.
+pub struct DataStream<R> {
```
yoks: i'm not sure if this is right place for it, maybe different file? bmc.rs holds just trait and adding structs here specific only for one of the functions looks out of place

core/src/bmc.rs
```diff
@@ -146,6 +186,19 @@ pub trait Bmc: Send + Sync {
         params: &T,
     ) -> impl Future<Output = Result<ModificationResponse<R>, Self::Error>> + Send;
 
+    /// POST a Redfish `UpdateService` multipart upload using a named stream.
+    fn multipart_update<U, V, R>(
```
yoks, line 153: Can you please make sure this signature does conforms Redfish spec about multipart https://www.dmtf.org/sites/default/files/standards/documents/DSP0266_1.23.0.html

README.md
```diff
@@ -126,6 +126,15 @@ See `examples/readme-minimal` for this example as a workspace target.
 See `examples/session-token` for Redfish SessionService authentication using
 `X-Auth-Token`.
 
+## Update Uploads
```
yoks: Maybe name it Multipart Uploads? Looks closer to spec, also not sure this part of Readme is where it should ist. But i have no strong opinion here.

core/src/bmc.rs
jayzhudev: Other than the upload_timeout, we are not supporting the OEM parts that the spec allows yet. Otherwise it conforms.

OEM specific parts | Content-Disposition | form-data; name="OemXXXX"
-- | -- | --

README.md
jayzhudev: Agreed, could be a left-over from my previous version, removing the whole section from README.

bmc-http/src/reqwest.rs
```diff
@@ -681,6 +700,51 @@ impl HttpClient for Client {
...
+            None => Part::stream(body),
+        };
+
+        let file_part = file_part
+            .file_name(name)
+            .mime_str("application/octet-stream")?;
+
+        let update_parameters_part = update_parameters_part(update_parameters)?;
+        let form = Form::new()
+            .part("UpdateParameters", update_parameters_part)
```
yoks, line 732: I think we need to support optional parts of form, like Oem data

bmc-http/src/lib.rs
```diff
@@ -112,6 +117,24 @@ pub trait HttpClient: Send + Sync {
         B: Serialize + Send + Sync,
         T: DeserializeOwned + Send + Sync;
 
+    /// Performs an UpdateService multipart upload with credentials and headers.
+    ///
+    /// Accepts a named stream so callers can upload firmware from any source
+    /// and still send Redfish's required `UpdateFile` multipart part.
+    fn post_multipart_update<U, V, T>(
```
yoks, line 125: This feels bloated in terms of amount of params, maybe we can reuse post (wrap body), or at least move multipart related data into separate struct.

core/src/bmc.rs
yoks: Yea, i would add them here, as optional.

bmc-http/src/lib.rs
jayzhudev: Yes, I used DataStream to wrap some but it does still look borderline too many. Let me think about it.

bmc-http/src/reqwest.rs
jayzhudev: Makes sense, will address this together with signature conformation to the spec, we will likely need the OEM parts in use cases.

core/src/bmc.rs
jayzhudev: I thought about it too. Let me try to find a better place for it.

bmc-http/src/lib.rs
jayzhudev: Added `MultipartUpdateRequest` to help with too many positional params:

```rust
/// Multipart `UpdateService` upload request data.
pub struct MultipartUpdateRequest<'a, U, V> {
    /// Redfish `UpdateParameters` JSON part.
    pub update_parameters: &'a V,

    /// Named stream sent as the Redfish `UpdateFile` part.
    pub update_stream: DataStream<U>,

    /// Optional extra multipart parts, such as OEM-defined parts.
    pub optional_parts: Vec<MultipartPart>,

    /// Timeout used only for this upload request.
    pub upload_timeout: Duration,
}
```

core/src/bmc.rs
jayzhudev: Created `core/src/upload.rs` for this as well as other multipart upload related error and structs.

core/src/bmc.rs
jayzhudev: Added with "Oem" name validation according to the spec.

> Optional OEM part. The value of the name field shall start with "Oem. Content-Type is optional and depends on the OEM part type.

redfish/src/update_service/mod.rs
```diff
@@ -278,3 +346,197 @@ fn add_default_update_service_name(v: JsonValue) -> JsonValue {
...
+    use std::error::Error as StdError;
+    use std::io;
+
+    const UPDATE_SERVICE_URI: &str = "/redfish/v1/UpdateService";
+    const MULTIPART_URI: &str = "/redfish/v1/UpdateService/update-multipart";
+
+    type TestResult = Result<(), Box<dyn StdError>>;
+
+    #[tokio::test]
+    async fn uses_multipart_http_push_uri() -> TestResult {
```
poroh: Could you please move these test to:
https://github.com/NVIDIA/nv-redfish/blob/main/tests/tests/test-update-service.rs
?

poroh (commented):
One small nit to move tests to other place. Everything else looks good to me to be merged.

redfish/src/update_service/mod.rs
jayzhudev: Good call, thanks! Moved these tests in the latest commit.

poroh (approved):

PR #103, opened 2026-05-21
fix(bmc-http): support typed mutation response bodies

jayzhudev:
Deserialize non-empty successful POST/PATCH bodies into the caller-selected response type even when the body has no `@odata.id`. This matches DSP0266 7.11, Table 10: action response bodies conform to their schema and are not necessarily Redfish resources.

Keep empty success responses as Empty, Redfish success envelopes as Empty for no-response actions, and 202 Location task handling unchanged.

### DIFF
--- a/bmc-http/src/reqwest.rs
+++ b/bmc-http/src/reqwest.rs
@@ -23,6 +23,7 @@ use crate::BmcCredentials;
 use crate::CacheableError;
 use crate::HttpClient;
 use crate::MultipartUpdateRequest;
+use crate::schema::redfish::redfish_error::RedfishError;
 
 use futures_util::StreamExt as _;
 use http::HeaderMap;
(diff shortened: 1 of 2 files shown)


yoks:
This one tricky to get right as we are pretty generic over resposne in the lib regardless if it action/create.

bmc-http/src/reqwest.rs
```diff
@@ -582,6 +603,28 @@ fn inject_etag(etag: &ODataETag, body: &mut serde_json::Value) {
...
+fn is_redfish_success_response(value: &serde_json::Value) -> bool {
+    // DSP0266 7.11, Table 10 allows actions without response bodies to return
+    // an error-shaped success body. Only that body should become Empty.
+    value
+        .get("error")
+        .and_then(serde_json::Value::as_object)
+        .and_then(|error| error.get("code"))
+        .and_then(serde_json::Value::as_str)
+        .and_then(|code| code.rsplit('.').next())
+        .is_some_and(|message| matches!(message, "Success" | "Created" | "NoOperation"))
```
yoks: Maybe there is better way than matching on content string?

yoks: Return types could be typed in CSDL, so maybe we can try to parse them first.

jayzhudev: Yeah this felt pretty ugly to me as well. Let me think a bit.

The response side is indeed tricky. The main goal is to stop dropping valid successful JSON responses when they don't include odata.id. So there's some room to make it less prescriptive.

yoks:
I know maybe i asking too much from this PR but there is https://redfish.dmtf.org/schemas/v1/RedfishError_v1.xml, we can try to reuse it. But problem is http crate does not contain this schema. So correct way is buble this error from the Redfish itself (which would contain schema). 

We can merge it as is for now. And create issue which describes Error handling in more details and uses CSDL schema.

jayzhudev:
> I know maybe i asking too much from this PR but there is https://redfish.dmtf.org/schemas/v1/RedfishError_v1.xml, we can try to reuse it. But problem is http crate does not contain this schema. So correct way is buble this error from the Redfish itself (which would contain schema).
> 
> We can merge it as is for now. And create issue which describes Error handling in more details and uses CSDL schema.

Exactly the same thought I had! Just created CSDL compiler PR https://github.com/NVIDIA/nv-redfish/pull/104 Once that's in, I'll update this PR to use it.

yoks (approved):

PR #107, opened 2026-05-23
feat(task-service)!: add spec compliant task service support

jayzhudev:
Adding this support for RMS firmware upload status checks.

Breaking change:

- When the `task-service` feature is enabled, the redfish::Error<B> now has 2 new variants.


redfish/src/task_service/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+    ///
+    /// Use this when an asynchronous response already returned a task path.
+    /// The path should be in the form `/redfish/v1/TaskService/Tasks/{id}`.
+    /// The returned generated schema exposes fields such as `task_state`,
+    /// `task_status`, `percent_complete`, and `messages`.
+    ///
+    /// # Errors
+    ///
+    /// Returns error if retrieving task data fails.
+    pub async fn task(&self, task_path: impl ToString) -> Result<Arc<Task>, Error<B>> {
```
poroh: If I understand correctly then it is from Location header from the result of the operation. It would be nice to:
1. Introduce type for a Location header value
2. Use it here and task service should check that Location header value corresponds to its own odata.id.

This will express additional knowledge trough type system and better errors if somebody by mistake will use task service for non-task values (I think some vendors may have their own async job services, see Dell OEM jobs for example).

poroh: Maybe we also should add separate function that can check that Location is valid Task resource.

jayzhudev: This is a very good and thoughtful call, thanks!

After some thinking, I see two options here:

1. add a Location type. The intent is clearer to the caller but it wouldn't prevent caller mistakes of calling a wrong task service instance - we would still need validation in `task` calling.

2. keep the API simple and only add validation inside TaskService::task(). The caller passes the task path, and TaskService checks it is under the BMC's Tasks collection before polling.

Option 2 seems to have the right tradeoff to me because the validation is task-specific, and TaskService already has the Tasks collection needed to do it.

redfish/src/task_service/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+///     return Ok(());
+/// };
+///
+/// let task = task_service.task(&async_task.id).await?;
+///
+/// println!("{:?}", task.task_state);
+/// ```
+pub struct TaskService<B: Bmc> {
+    data: Arc<TaskServiceSchema>,
+    task_collection: TaskCollectionPath,
```
poroh: Think we should not keep this information inside TaskService. All information is in data (odata.id). Just add method to ODataId type that can check that one OdataId `is_path_prefix` for another. No need to `trim_end_matheces` etc. Just add one method here and use it is needed.

https://github.com/NVIDIA/nv-redfish/blob/b5dc7846ec850e044a9d817389612cb3a15923f0/core/src/odata.rs#L51

redfish/src/task_service/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+use crate::ResourceSchema;
+use crate::ServiceRoot;
+use crate::core::Bmc;
+use crate::core::EntityTypeRef as _;
+use crate::core::NavProperty;
+use crate::core::ODataId;
+use crate::schema::task_service::TaskService as TaskServiceSchema;
+
+#[doc(inline)]
+pub use crate::schema::task::Task;
```
poroh: nv-redfish has two concept:
1. Full wrapper like Chassis where you have different custom methods 
2. Link that internally contains just odata-id reference for the object.

It seems that Task should be a link here:



```suggestion
use crate::entity_link::EntityLink;
use crate::schema::task::Task as TaskSchema;

/// Link for accessing task.
pub type TaskLink<B> = EntityLink<B, TaskSchema>;
```

redfish/src/task_service/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+    /// The path must be under this service's Tasks collection, such as
+    /// `/redfish/v1/TaskService/Tasks/{id}`.
+    /// The returned generated schema exposes fields such as `task_state`,
+    /// `task_status`, `percent_complete`, and `messages`.
+    ///
+    /// # Errors
+    ///
+    /// Returns error if the service does not expose a Tasks collection, if the
+    /// path is outside that collection, or if retrieving task data fails.
+    pub async fn task(&self, task_path: impl AsRef<str>) -> Result<Arc<Task>, Error<B>> {
```
poroh:
```
use crate::core::AsyncTask;
```

```suggestion
    pub fn task_link(&self, async_task: AsyncTask) -> Result<TaskLink, Error<B>> {
```
Internally task_link should check that task service odata id is parent path for AsyncTask::id and constructs link only if this is true.

Note that this method doesn't retrieve any data, just creates handle that can be used to retrieve data. See:
https://github.com/NVIDIA/nv-redfish/blob/b5dc7846ec850e044a9d817389612cb3a15923f0/redfish/src/entity_link.rs#L94

poroh (changes requested):

redfish/src/task_service/mod.rs
jayzhudev: Thanks for sharing this! This utility seems to be the right pattern to use. I've updated the implementation.

redfish/src/task_service/mod.rs
jayzhudev: Agreed, I've updated the implementation.

redfish/src/task_service/mod.rs
```diff
@@ -49,21 +52,17 @@ pub use crate::schema::task::Task;
...
-/// let task = task_service.task(&async_task.id).await?;
+/// let task_link = task_service.task_link(&async_task)?;
+/// let task = task_link.fetch().await?;
 ///
 /// println!("{:?}", task.task_state);
 /// ```
 pub struct TaskService<B: Bmc> {
     data: Arc<TaskServiceSchema>,
-    task_collection: TaskCollectionPath,
+    task_collection: ODataId,
```
poroh: This copy of odata_id is redundant. You can use data.odata_id() when this value is needed.

redfish/src/task_service/mod.rs
```diff
@@ -102,28 +95,27 @@ impl<B: Bmc> TaskService<B> {
...
     /// # Errors
     ///
-    /// Returns error if the service does not expose a Tasks collection, if the
-    /// path is outside that collection, or if retrieving task data fails.
-    pub async fn task(&self, task_path: impl AsRef<str>) -> Result<Arc<Task>, Error<B>> {
-        let task_path = task_path.as_ref();
-        if !task_path.starts_with(self.task_collection.prefix.as_str()) {
+    /// Returns error if the task ID is not a child of this service's Tasks
+    /// collection.
+    pub fn task_link(&self, task: &AsyncTask) -> Result<TaskLink<B>, Error<B>> {
```
poroh: To avoid "clones" just use `task: AsyncTask`.

jayzhudev: The tradeoff is around consuming the task versus cloning just the id (current implementation). I chose the latter in case the caller still needs the task (for the retry_after_secs field).

redfish/src/task_service/mod.rs
jayzhudev: Replaced it with data.tasks.odata_id.

poroh (approved):

PR #108, opened 2026-05-26
feat(control): add Redfish Control support for power limits

jayzhudev:
This adds typed support for the standard Redfish Control schema and resolves `EnvironmentMetrics.PowerLimitWatts` to its target Control resource.

RMS needs to `PATCH` power limits, and Redfish represents that power-limit target as a Control resource. Adding this support lets nv-redfish follow the schema directly instead of handling the power-limit URI as a special case.


yoks (approved):
Looks good, only nit is, we are seem to be incosistent with expand vs get sematnic, some collection uses expand and some not. We should go trhough codebase and standartise it.

poroh (approved):

PR #109, opened 2026-05-31
feat(upload): support legacy HttpPushUri uploads

jayzhudev:
Added opt-in support for `UpdateService.HttpPushUri`, the older Redfish firmware upload path.

`HttpPushUri` is deprecated in favor of `MultipartHttpPushUri`, and multipart should stay the preferred path. But some existing RMS firmware update flows still use the raw binary upload endpoint.

This addition is put behind `update-service-deprecated` feature which lets callers only get this API when they explicitly opt in. Same pattern could be used for managing future spec deprecations.

Spec references:
- DMTF marks `HttpPushUri`, targets, options, and busy flags deprecated in `UpdateService` v1.15.0 in favor of `MultipartHttpPushUri`.
- DSP0266 defines the multipart push path and common mutation/task response behavior.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,6 +69,11 @@ define build-and-test
 	cargo build -p nv-redfish-tests --tests
 	cargo build -p nv-redfish-bmc-mock
 	cargo test $1 -- --no-capture
+	cargo build -p nv-redfish --features update-service-deprecated
+	cargo build -p nv-redfish --features bmc-http,update-service-deprecated
+	cargo test -p nv-redfish-bmc-http --test reqwest_client_tests --features reqwest,update-service-deprecated
+	cargo test -p nv-redfish-tests --test test-update-service --features update-service-deprecated
+	cargo build -p update-multipart --features update-service-deprecated
 	cargo clippy $1
 	cargo build  $1
 	cargo build -p nv-redfish --features computer-systems,bios,boot-options,storages,memory,processors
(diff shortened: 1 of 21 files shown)


yoks:
Not sure why we need first class support for depricated upload API?

I would assume if BMC/Endpoint does not support Multipart we should transperently falback to old upload?

jayzhudev:
> Not sure why we need first class support for depricated upload API?
> 
> I would assume if BMC/Endpoint does not support Multipart we should transperently falback to old upload?

Preferring the explicit opt-in instead of changing the semantics of multipart update for:

- caller intent is explicit and there's clearer reasoning about failures
- caller controls fallback implementation and logic without having to understand or work around the nv-redfish fallback logic -> more flexibility and clarity
- "UpdateParameters" may be implicitly lost on the fallback path (the signatures are different)

We could leave the fallback implementation to the caller if the BMC's capability is unknown, or add a discovery time helper if we want to.

yoks:
I would aggree, hiding logic is missleading. I do not like ergonomics of that call (e.g. forcing user to explicitly discover which upload works), but arguably this is better than hidden fallback.

As minor nits, there is difference in name for DataStream in case of Upload (filename) and Multipart. And i think if we want to set deprecated/sunset in the crates consistent, we need to mark every deprecated path with feature flag, including core.

jayzhudev:
> I would aggree, hiding logic is missleading. I do not like ergonomics of that call (e.g. forcing user to explicitly discover which upload works), but arguably this is better than hidden fallback.
> 
> As minor nits, there is difference in name for DataStream in case of Upload (filename) and Multipart. And i think if we want to set deprication/sunset in the crates consistent, we need to mark every depricated path with feature flag, including core.

Thanks! Agreed, will improve the consistency here in the next commit.

poroh (approved):

PR #110, opened 2026-06-05
feat(bmc-http): Add configurable retry policy to reqwest client

kurotych:
## Description
Add `RetryPolicy` to `ClientParams` with a user-provided classifier deciding which (request, response) pairs to retry, a configurable `max_retries`, and an optional fixed `delay`. Streaming bodies (multipart uploads) are never retried since they cannot be cloned.

Routing everything through `send()` keeps a single code path for executing requests. It's safe for streaming because the retry loop self-disables there: try_clone() returns None for streaming bodies.  
  
## Motivation

We have a case where a BMC sporadically responds with 403 Forbidden to (any) GET request during exploration. These responses come from the lighttpd frontend (it returns the HTTP header `Server: lighttpd`). The only workaround (I found experimentally) is to sleep 8 seconds and retry the same request.

## Solution
Since described case is quite unique, it is proposed adding a retry mechanism where the decision logic for whether an error is retryable is fully configurable from the client side. It allows to cover such cases without patches or forks. 

## Alternatives
reqwest 0.12 ships a retry module, but it doesn't fit because it still doesn't have delay between attempts. 

## (NICo) Usage example

```rust
let retry_policy = RetryPolicy::new(|request, response| {
    *request.method() == http::Method::GET
        && response.status() == reqwest::StatusCode::FORBIDDEN
        && response
            .headers()
            .get(reqwest::header::SERVER)
            .and_then(|v| v.to_str().ok())
            .is_some_and(|v| v.eq_ignore_ascii_case("lighttpd"))
})
.max_retries(3)
.delay(Duration::from_secs(8));

let client = RedfishReqwestClient::with_params(
    RedfishReqwestClientParams::new()
        .accept_invalid_certs(true)
        .retry(retry_policy),
)
.map_err(|err| Error::Bmc(err.into()))?;
...
```

### DIFF
--- a/bmc-http/Cargo.toml
+++ b/bmc-http/Cargo.toml
@@ -14,7 +14,7 @@ documentation = "https://docs.rs/nv-redfish-bmc-http"
 default = ["reqwest"]
 
 # HTTP client implementations with reqwest
-reqwest = ["dep:reqwest", "dep:serde_path_to_error", "dep:futures-util", "dep:sse-stream", "dep:tokio-util"]
+reqwest = ["dep:reqwest", "dep:serde_path_to_error", "dep:futures-util", "dep:sse-stream", "dep:tokio-util", "dep:tokio"]
 update-service-deprecated = ["nv-redfish-core/update-service-deprecated"]
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


yoks:
looks clean, but i would maybe not do it for all requests. As it makes all request retry-able. Meaning POST/PUT (which is most cases are not idempotent).

So i would either document it somewhere on RetryPolicy docs or add additional parameter to retry all request types or just do it only for Get.

kurotych:
Thanks for review @yoks 

I've extended the description and added a test (showcase) demonstrating
that other HTTP methods are not retried when the `RetryClassifier` is configured properly.

Hope it is clearer now.

yoks (approved):

PR #112, opened 2026-06-13
feat(bmc-http)!: resolve action targets as URI references

jayzhudev:
I debated this a bit, but this seems like a good addition overall. It lets `nv-redfish` handle action target URLs the same way it already handles update and stream URLs. If the service returns an absolute action target, we now use that URL directly instead of forcing it under the configured BMC endpoint.

The main tradeoff is that absolute action targets are followed as provided by the service. If such a target uses a different host than the configured BMC endpoint, the action request goes to that host using the `HttpBmc` credentials and custom headers. This flexibility already exists elsewhere in the client, but this extends it to actions. It seems reasonable to me, since outbound policy should be the caller's responsibility. I am open to feedback if a stricter default would be preferred.

### DIFF
--- a/bmc-http/src/lib.rs
+++ b/bmc-http/src/lib.rs
@@ -408,11 +408,11 @@ impl RedfishEndpoint {
     fn with_uri_reference(&self, uri: UriReference<'_>) -> Url {
         let UriReference(uri) = uri;
 
-        let Ok(url) = self.base_url.join(uri) else {
+        let Ok(resolved) = self.base_url.join(uri) else {
             return self.with_path(uri);
         };
 
-        url
+        resolved
     }
 
     /// Convert a path to a full Redfish endpoint URL with query parameters
--- a/bmc-http/src/lib.rs
+++ b/bmc-http/src/lib.rs
@@ -199,6 +199,13 @@ pub trait HttpClient: Send + Sync {
 /// to provide a complete Redfish client implementation. It implements the [`Bmc`] trait
 /// to provide standardized access to Redfish services.
 ///
+/// For Redfish URI-reference fields, such as action targets and update or
+/// stream URIs, this implementation resolves the URI reference and sends the
+/// request with the configured credentials and custom headers. It does not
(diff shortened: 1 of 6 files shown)


poroh:
Is it real scenario when some BMC provides full URI for action? It sounds kind of dangerous to me to support this behavior. In general, I would prefer to restrict actions and uploads to local paths only and enforce it on action / uploads level. If full URI is specified as target I would just give possibility to users to retrieve it and explicitly call requests on their own and decide whether it safe to provide credentials / tokens to the target.

jayzhudev:
> Is it real scenario when some BMC provides full URI for action? It sounds kind of dangerous to me to support this behavior. In general, I would prefer to restrict actions and uploads to local paths only and enforce it on action / uploads level. If full URI is specified as target I would just give possibility to users to retrieve it and explicitly call requests on their own and decide whether it safe to provide credentials / tokens to the target.

Yeah, that's the tradeoff I was debating and I agree with this concern.

I have not seen a full URI action target from our test equipment, but that sample size is relatively small.

The spec allows an absolute URL in section 9.6.14.1: https://www.dmtf.org/sites/default/files/standards/documents/DSP0266_1.23.0.html#task 

> The property for the action is a JSON object and contains the following properties:
>
> The target property shall be present and defines the relative **or absolute URL** to invoke the action.
> The title property may be present and defines the action's name.

A middle ground could be adding same-origin verification: `nv-redfish` can accept relative and same-origin absolute targets, but reject cross-origin targets before sending credentials or custom headers. Callers can still inspect those targets and handle them explicitly if they want to.

jayzhudev:
@poroh Thanks for the review. I added validation so cross-origin or malformed Redfish URI references are rejected before sending credentials or custom headers.

Note: after 751448857e357ea9e5f9c425983dc66878dcf165, this is source-breaking for custom `HttpClient` users because their error type now needs to implement `RequestError`.

No need to force this change though. I’m open to closing the PR if there are still security concerns.

poroh (approved):

yoks (approved):

PR #113, opened 2026-06-13
Handle zero-capacity cache safely

fallintoplace:
## Summary

Treat a cache capacity of 0 as an explicitly disabled cache instead of entering CAR replacement on the first insert.

## Details

`CacheSettings::with_capacity(0)` can currently build a `CarCache` with empty clock lists. On the first cache miss, `put` sees `len == capacity`, calls `replace`, and the clock hand can advance with modulo zero.

This keeps the existing `usize` API intact and makes the compatible behavior explicit: capacity 0 stores no entries and returns cache misses.

## Validation

- `cargo test -p nv-redfish-bmc-http`

### DIFF
--- a/bmc-http/src/cache.rs
+++ b/bmc-http/src/cache.rs
@@ -293,7 +293,9 @@ impl<K, V> CarCache<K, V>
 where
     K: Eq + Hash + Clone,
 {
-    /// Create new CAR cache with given capacity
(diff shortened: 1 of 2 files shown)


yoks:
Can you plase sign DCO on your commits? Thanks

yoks (approved):

PR #114, opened 2026-06-19
feat(redfish): expose standard reset action helpers

jayzhudev:
Add typed high-level wrappers for DMTF Redfish reset actions on `ComputerSystem`, `Manager`, `Chassis`, and `PowerSupply`.

Spec references:
- ComputerSystem.Reset: schema/redfish-csdl/csdl/ComputerSystem_v1.xml
- Manager.Reset: schema/redfish-csdl/csdl/Manager_v1.xml
- Manager.ResetToDefaults: schema/redfish-csdl/csdl/Manager_v1.xml
- Chassis.Reset: schema/redfish-csdl/csdl/Chassis_v1.xml
- PowerSupply.Reset: schema/redfish-csdl/csdl/PowerSupply_v1.xml

Separate csdl-compiler fixes have been moved to PR https://github.com/NVIDIA/nv-redfish/pull/115

### DIFF
--- a/redfish/src/chassis/item.rs
+++ b/redfish/src/chassis/item.rs
@@ -23,12 +23,14 @@ use crate::hardware_id::SerialNumber as HardwareIdSerialNumber;
 use crate::patch_support::JsonValue;
 use crate::patch_support::Payload;
 use crate::patch_support::ReadPatchFn;
+use crate::resource::ResetType;
 use crate::schema::chassis::Chassis as ChassisSchema;
 use crate::Error;
 use crate::NvBmc;
 use crate::Resource;
 use crate::ResourceSchema;
 use nv_redfish_core::bmc::Bmc;
+use nv_redfish_core::ModificationResponse;
 use nv_redfish_core::NavProperty;
 use std::future::Future;
 use std::sync::Arc;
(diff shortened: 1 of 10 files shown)


csdl-compiler/src/generator/rust/struct_def.rs
```diff
@@ -706,34 +706,44 @@ impl<'a> StructDef<'a> {
...
                     &e,
                     quote! { #top::Reference },
-                    rename,
                     p.nullable,
                     p.required,
                     RigidArraySupport::new(false),
                 )
             }
         };
+        let serde = Self::gen_action_parameter_serde_annotation(rename, p.required);
```
poroh: I have concerns WRT this function. gen_de_struct_field_type can generate , `T`, `Option<T>`, `Option<Option<T>`,  `Vec<T>`, `Option<Vec<T>`, `Option<Option<Vec<T>`. gen_action_parameter_serde_annotation only adds `"skip_serializing_if = Option::is_none"`.

Maybe old generated code was not quite correct in some situations however I don't think that new code can correctly handle "null" fields. 

I think we need something like:
```
    fn gen_ser_struct_field_serde_annot(
        rename: impl ToTokens,
        nullable: IsNullable,
        required: IsRequired,
    ) -> TokenStream {
        if required.into_inner() && nullable.into_inner() {
            quote! { #[serde(rename=#rename, serialize_with="ser_required_nullable")] }
        } else if required.into_inner() {
            quote! { #[serde(rename=#rename)] }
        } else if nullable.into_inner() {
            quote! { #[serde(rename=#rename, default, serialize_with="ser_optional_nullable")] }
        } else {
            quote! { #[serde(rename=#rename, default, skip_serializing_if = "Option::is_none"))] }
        }
    }
```

poroh (changes requested):

csdl-compiler/src/generator/rust/struct_def.rs
jayzhudev: That's a good point. Functionality wise, the `gen_de_struct_field_type` decides the type such as the ones you mentioned (`Option<T>`, `Option<Option<T>>` etc.). So the nullability is "encoded" by it. The `gen_action_parameter_serde_annotation ` controls if the top-level `None` should be omitted. One concern is to potentially having a `Vec<T>` field annotated with `skip_serializing_if = "Option::is_none"`. 

The key non-obviousness comes from implicitly coordinating two branching logic based on `required`. In `gen_de_struct_field_type`, `required` generates a non-`Option`, or an `Option` that can be serialized to JSON `null`, which matches the branches in this function.

I added this clarification in comments and added tests to catch regression.

jayzhudev: Maybe we can think of some refactoring to make this coordination obvious?

poroh (approved):

PR #116, opened 2026-06-21
Fix query-aware HTTP cache keys

fallintoplace:
## Summary
- key HTTP cache and ETag entries by the full request URL instead of only the ODataId
- keep different $expand and $filter representations independent even when they share an ETag
- add regression tests for distinct query responses and 304 cache hits

## Validation
- cargo fmt -p nv-redfish-bmc-http -- --check
- cargo test -p nv-redfish-bmc-http --features reqwest

### DIFF
--- a/bmc-http/src/lib.rs
+++ b/bmc-http/src/lib.rs
@@ -214,8 +214,8 @@ pub struct HttpBmc<C: HttpClient> {
     client: C,
     redfish_endpoint: RedfishEndpoint,
     credentials: RwLock<Arc<BmcCredentials>>,
-    cache: RwLock<TypeErasedCarCache<ODataId>>,
-    etags: RwLock<HashMap<ODataId, ODataETag>>,
+    cache: RwLock<TypeErasedCarCache<Url>>,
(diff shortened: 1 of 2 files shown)


yoks:
I see why we want url, but does expand and filter really utilize etag anywhere? Never met this in wild.

Also please sign DCO, without it we cannot merge it.

fallintoplace:
Thanks! DCO should be fixed now. @yoks 

On the ETag bit: agreed, I also don’t expect this to be common in the wild. The reason I still think the URL key is the right fix is that this client does not treat queried reads differently from plain `get`: `expand` and `filter` both go through the same `get_with_cache` path, and if the response has an ETag we’ll send `If-None-Match` on the next request.

So the bug is mostly defensive/client-side correctness: once a queried response is cacheable, the representation key needs to include the query string. Otherwise different `$expand`/`$filter` representations for the same `@odata.id` can overwrite each other or serve the wrong cached body on a 304. The tests are synthetic for that edge case, not a claim that most BMCs do this today.

yoks (approved):

PR #117, opened 2026-06-21
Fix pre-epoch fractional SystemTime conversion

fallintoplace:
## Summary
- fix EdmDateTimeOffset to SystemTime conversion for fractional timestamps before the Unix epoch
- borrow from the seconds field before subtracting negative fractional timestamps from UNIX_EPOCH
- add a regression test for 1969-12-31T23:59:59.5Z

## Validation
- cargo fmt -p nv-redfish-core -- --check
- cargo test -p nv-redfish-core


yoks:
Same for DCO here. Thanks!

yoks (approved):

PR #118, opened 2026-06-23
fix: add bmc-quirks for lenovo-gb300 [AMI]

krish-nvidia:
## Summary
AMI host BMCs return `$expand` collection responses that drop Redfish-`Required` fields (`Id`, `Name`, `ChassisType`) from embedded members, causing deserialization to fail with `missing field ...` for the whole collection. The standalone resource GETs are complete however.

We already handle this for Viking via `expand_is_not_working_properly`, which makes nv-redfish fetch each member with an individual GET. This extends the same workaround to the Grace-based NVIDIA GB300 host BMC, identified by its AMI OEM `RtpVersion` in the service root.

## Changes
- Add a `Platform::AmiGb300` variant.
- Classify `Vendor == "AMI"` with `Oem.Ami.RtpVersion == "13.09.1"` (and not Viking) as `Platform::AmiGb300`. Other AMI BMCs are unchanged and keep using `$expand`.
- Extend `expand_is_not_working_properly()` to `AmiViking | AmiGb300`.

`Platform::AmiGb300` deliberately does **not** inherit Viking's other quirks (the `filter_computer_system_odata_ids` / `filter_manager_odata_ids` allowlists are Viking-topology-specific and would drop GB300's real host system/manager; the `ChassisType` / `Name` / root-nav defaults aren't needed since GB300's standalone GETs are complete).

## Testing
Added integration tests (`test-chassis.rs`): a GB300 root (`RtpVersion 13.09.1`) fetches members individually, and a non-GB300 AMI root still uses `$expand`.

Verified against a live Lenovo GB300 BMC: exploration previously failed on `missing field ChassisType` / `Id`; with expand disabled it now returns a full report (chassis, systems, managers, firmware inventory).

Even with expand functionality removed, report generation happens relatively quickly:
```
real    0m21.550s
user    0m0.784s
sys     0m0.188s
```

### DIFF
--- a/redfish/src/bmc_quirks.rs
+++ b/redfish/src/bmc_quirks.rs
@@ -32,7 +32,7 @@ enum Platform {
     Hpe,
     Dell,
     AmiViking,
-    Ami,
+    AmiGb300,
     Nvidia,
     NvidiaDpu,
     Anonymous1_9_0,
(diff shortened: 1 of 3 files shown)


redfish/src/bmc_quirks.rs
```diff
@@ -47,6 +48,7 @@ impl BmcQuirks {
             Some("HPE") => Some(Platform::Hpe),
             Some("Dell") => Some(Platform::Dell),
             Some("AMI") if redfish_version_str == Some("1.11.0") => Some(Platform::AmiViking),
+            Some("AMI") => Some(Platform::Ami),
```
poroh: It looks bit broad to disable expand for all AMIs. Can we narrow down it somehow?

poroh (approved):

PR #121, opened 2026-06-28
fix(bmc-http): disable ETags when cache capacity is zero

jayzhudev:
Cache capacity zero disables caching, but the following workflow would fail:

- The first GET returns `200` with a response body and an ETag.
- `put_typed` does not store the response entity when cache capacity is zero, but the ETag is still stored in `etags`.
- A subsequent GET sends `If-None-Match` using that ETag.
- If the BMC supports `If-None-Match` and the resource is unchanged, it returns `304 Not Modified` without a response body [RFC 9110 15.4.5](https://www.rfc-editor.org/rfc/rfc9110.html#section-15.4.5).
- `HttpBmc` attempts to load the response entity from its zero-capacity cache.
- The caller receives `CacheMiss`.

Changed resources would still work in this case as `200` will be returned. Unchanged resources cause a `CacheMiss` when the BMC supports `If-None-Match` and returns `304`.

### DIFF
--- a/bmc-http/src/lib.rs
+++ b/bmc-http/src/lib.rs
@@ -568,15 +568,13 @@ where
             )
             .await
         {
-            Ok(response) => {
-                let entity = Arc::new(response);
-
+            Ok(response) if !self.cache_enabled => {
                 // With capacity zero, `put_typed` stores no representation and always returns
                 // `None`, and we can return early with the response entity.
-                if !self.cache_enabled {
-                    return Ok(entity);
-                }
-
+                Ok(Arc::new(response))
+            }
+            Ok(response) => {
+                let entity = Arc::new(response);
                 // Update cache if entity has etag
                 if let Some(etag) = entity.etag() {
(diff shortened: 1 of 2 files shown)


bmc-http/src/lib.rs
```diff
@@ -560,6 +571,12 @@ where
             Ok(response) => {
                 let entity = Arc::new(response);
 
+                // With capacity zero, `put_typed` stores no representation and always returns
+                // `None`, and we can return early with the response entity.
+                if !self.cache_enabled {
+                    return Ok(entity);
+                }
+
```
poroh:
```suggestion
            Ok(response) if !self.cache_enabled => {
                // With capacity zero, `put_typed` stores no representation and always returns
                // `None`, and we can return early with the response entity.
                Ok(Arc::new(response))
            }
            Ok(response) => {
                let entity = Arc::new(response);
```

poroh (approved):

PR #126, opened 2026-07-01
fix(core): prevent empty OData type panic

jayzhudev:
Fix a panic in `ODataType::parse_from` when BMC JSON contains an empty `@odata.type`.

## Root cause

```rust
.and_then(|v| v.starts_with('#').then_some(&v[1..]))
```

`bool::then_some` evaluates its argument eagerly. For an empty string, `&v[1..]` is evaluated before the `starts_with('#')` result is applied, causing an out-of-bounds panic.

## Minimal reproduction

```rust
use serde_json::json;

fn main() {
    let s = json!({ "@odata.type": "" });
    let _ = nv_redfish_core::odata::ODataType::parse_from(&s);
}
```

### DIFF
--- a/core/src/odata.rs
+++ b/core/src/odata.rs
@@ -136,7 +136,7 @@ impl ODataType<'_> {
     pub fn parse_from(v: &serde_json::Value) -> Option<ODataType<'_>> {
         v.get("@odata.type")
             .and_then(|v| v.as_str())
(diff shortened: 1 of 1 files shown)


yoks:
Any live example of such behavior?

yoks (approved):

jayzhudev:
> Any live example of such behavior?

I'm not aware of a real BMC causing such issue but it seems to be a good hardening 😃

PR #127, opened 2026-07-03
fix(account)!: preserve modify responses

jayzhudev:
## Why

Redfish account changes can finish with an entity, finish with no body, or return a task. The `Option<Account>` API turned both `Task` and `Empty` into `None`, losing the task handle callers need to track completion and failures.

## Changes

- Return `ModificationResponse<Account<B>>` from account create/update/delete paths.
- Preserve `Entity`, `Task`, and `Empty` instead of collapsing non-entity success.
- Add mock support and integration coverage for task and empty responses.

## Breaking Change

Account create/update/delete APIs now return `ModificationResponse<Account<B>>` instead of `Option<Account<B>>`, so callers need to handle `Entity`, `Task`, and `Empty`.

### DIFF
--- a/redfish/src/account/item.rs
+++ b/redfish/src/account/item.rs
@@ -118,32 +118,49 @@ impl<B: Bmc> Account<B> {
 
     /// Update the account.
     ///
-    /// Returns the new (updated) account.
+    /// Returns one of the following modification outcomes:
+    ///
+    /// - `ModificationResponse::Entity` contains the updated account.
+    /// - `ModificationResponse::Task` identifies an asynchronous operation.
+    /// - `ModificationResponse::Empty` reports synchronous success without a
+    ///   response body.
     ///
     /// # Errors
     ///
     /// Returns an error if the server responds with an error or if the
     /// response cannot be parsed.
-    pub async fn update(&self, update: &ManagerAccountUpdate) -> Result<Option<Self>, Error<B>> {
+    pub async fn update(
+        &self,
+        update: &ManagerAccountUpdate,
+    ) -> Result<ModificationResponse<Self>, Error<B>> {
         match self.update_with_patch(update).await? {
-            ModificationResponse::Entity(ma) => Ok(Some(Self::from_data(
+            ModificationResponse::Entity(ma) => Ok(ModificationResponse::Entity(Self::from_data(
                 self.bmc.clone(),
                 ma,
                 self.config.clone(),
             ))),
-            ModificationResponse::Task(_) | ModificationResponse::Empty => Ok(None),
+            ModificationResponse::Task(task) => Ok(ModificationResponse::Task(task)),
+            ModificationResponse::Empty => Ok(ModificationResponse::Empty),
         }
     }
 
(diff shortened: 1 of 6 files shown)


redfish/src/account/item.rs
```diff
@@ -118,32 +118,49 @@ impl<B: Bmc> Account<B> {
...
     /// Returns an error if the server responds with an error or if the
     /// response cannot be parsed.
-    pub async fn update(&self, update: &ManagerAccountUpdate) -> Result<Option<Self>, Error<B>> {
+    pub async fn update(
+        &self,
+        update: &ManagerAccountUpdate,
+    ) -> Result<ModificationResponse<Self>, Error<B>> {
         match self.update_with_patch(update).await? {
-            ModificationResponse::Entity(ma) => Ok(Some(Self::from_data(
+            ModificationResponse::Entity(ma) => Ok(ModificationResponse::Entity(Self::from_data(
```
yoks: Is it equal to e => e?

jayzhudev: Ha! I think you are right, let me fix.

yoks:
Which Hardware exhibit this behavior?

redfish/src/account/item.rs
jayzhudev: This actually is needed or we need to add a map method because the following doesn't auto convert `ManagerAccount` to `Account<B>` because of `ModificationResponse<T>` T generics.

```rust
if let ... else { Ok(r) }
```

Or we can create a map method that takes `FnOnce` for ergonomics, but may or may not make it more clear. Personally I prefer this over a map method.

jayzhudev:
> Which Hardware exhibit this behavior?

This is a cleanup for spec compliance:

https://www.dmtf.org/sites/default/files/standards/documents/DSP0266_1.23.0.html#modification-success-responses

image

Basically it allows `202 Accepted` in response, in that case:

> The HTTP [202 Accepted](https://www.dmtf.org/sites/default/files/standards/documents/DSP0266_1.23.0.html#status-202) status code with a [Location header](https://www.dmtf.org/sites/default/files/standards/documents/DSP0266_1.23.0.html#location-header-table) set to the URI of a task monitor when the processing of the modification requires additional time.

So if the response `Task` is returned as `None`, the caller loses the task handle.

jayzhudev:
The involved methods are data modification requests so `202` responses are allowed -- this PR is a compatibility widening change.

redfish/src/account/item.rs
yoks: Maybe this wrapping just makes our life harder? Seems like return of ManagerAccount seems more straightforward, we already make them public.

yoks: Also, maybe be controversial, but map functions will make it alittle more DRY as it repeated in collection.rs as well, can be reused in both places?

jayzhudev: Yes, I like the `map` idea more than returning `ManagerAccount`. Maybe we can try that?

yoks (approved):

PR #130, opened 2026-07-06
Dispatcher Graceful Shutdown and Leaf/Bucket Schedulers

yoks:
Added graceful shutdown hanlding to dispatcher, as well as two new schdulers:

- FixedLeaf
- TokenBucket

Added ability to remove nodes from RoundRobin scheduler.

### DIFF
--- a/dispatcher/src/schedulers/token_bucket.rs
+++ b/dispatcher/src/schedulers/token_bucket.rs
@@ -1,0 +1,510 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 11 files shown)


poroh (approved):

dispatcher/src/schedulers/token_bucket.rs
```diff
@@ -0,0 +1,510 @@
...
+{
+    /// Create a new [`TokenBucket`] with the given config and child
+    /// scheduler. The bucket starts at full `capacity`.
+    ///
+    /// The child's meta must expose [`crate::HasCost`] so the actual
+    /// cost of each dispatched item can be charged; adapt cost-naive
+    /// children with [`crate::schedulers::FixedCost`].
+    #[must_use]
+    pub fn new(cfg: TokenBucketConfig, child: C) -> Self {
+        let now = Instant::now();
```
poroh: nit: It is better to pass now as parameter to make time handling purely provided by user.

dispatcher/src/schedulers/token_bucket.rs
```diff
@@ -0,0 +1,510 @@
...
+        if self.interval_nanos() == 0 {
+            // Unlimited rate: always full.
+            self.scaled_balance = cap;
+            self.last_refill = now;
+            return;
+        }
+        let elapsed = now.saturating_duration_since(self.last_refill);
+        self.last_refill = now;
+        let accrued = i128::try_from(elapsed.as_nanos()).unwrap_or(i128::MAX)
+            * i128::from(self.cfg.refill_amount.get());
```
poroh: nit: if elapsed is unwrapped to i128::MAX then it definitely will cause overflow on multiplication and will panic. I would use checked_mul instead of plain multiplication to avoid panics.

yoks: Agree, same can be done for PereodicLeaf

poroh (approved):

PR #135, opened 2026-07-10
CAR Cache correctness and performance improvements

yoks:
## CAR cache: performance, paper conformance, and benchmarks

With help of Claude Fable and GPT-5.6sol. Added gungraun perf benchmark and walked through instruction set. Main change is to make ClockList intrusive, which leads to good performance improvements on scans.

Also found various small (non-critical bugs), AI description bellow goes through them.

## AI Generated description bellow 

Reworked the `bmc-http` CAR cache hot paths and cross-checked the
implementation against the CAR paper (Bansal & Modha, FAST'04), fixing three
conformance bugs.

- **Perf**: `ClockList` is now an intrusive circular ring (O(1) head ops),
  index updates are clone-free, slot storage allocates lazily. Full-clock
  sweep −22% instructions, 256-key scan −83%.
- **Conformance**: fixed a T1 double hand-advance that skipped pages during
  replacement; write hits now set the reference bit (paper line 2); ghost
  discards happen only via the guarded directory rules — ghost lists carry
  one slack slot instead of auto-evicting, which could drop the requested
  key's own ghost mid-put and lose its adaptation hit.
- **Tests**: behavioral test for the slack slot (fails on old code), exact
  pins for both adaptation deltas, trajectory-shape asserts instead of
  brittle endpoint pins. Verified additionally by a 24M-op randomized fuzz.
- **Benchmarks/CI**: new gungraun cache bench suite; fixed
  `insert_with_headroom` (was measuring an eviction on a full half-size
  cache); CI regression job now gates both dispatcher and cache benches at
  `ir=2%`.

### DIFF
--- a/bmc-http/src/cache.rs
+++ b/bmc-http/src/cache.rs
@@ -18,7 +18,11 @@
 //! Based on "CAR: Clock with Adaptive Replacement" by Bansal & Modha
 //! USENIX Conference on File and Storage Technologies, 2004
 //!
-//! This implementation follows the exact pseudocode from the [paper](https://www.usenix.org/legacy/publications/library/proceedings/fast04/tech/full_papers/bansal/bansal.pdf).
+//! This implementation follows the pseudocode from the [paper](https://www.usenix.org/legacy/publications/library/proceedings/fast04/tech/full_papers/bansal/bansal.pdf),
+//! with one clock-native reading: "make it the tail page in T2"
+//! (line 36) is realized by clearing the reference bit and advancing
+//! the hand — in a circular list that leaves the page as the last one
+//! the hand revisits.
 
 use std::any::Any;
 use std::collections::HashMap;
@@ -64,7 +68,7 @@ impl<K, V> CacheEntry<K, V> {
 }
 
 /// Node in the ghost list doubly-linked structure
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 struct GhostNode<K> {
     key: K,
     prev: Option<usize>,
@@ -74,7 +78,10 @@ struct GhostNode<K> {
 /// Intrusive doubly linked list for ghost entries (B1, B2)
 #[derive(Debug)]
 struct GhostList<K> {
+    /// Grows on demand up to `capacity`, so an idle cache holds no
+    /// slot storage.
     entries: Vec<Option<GhostNode<K>>>,
+    capacity: usize,
     head: Option<usize>, // LRU end
     tail: Option<usize>, // MRU end
     free_slots: Vec<usize>,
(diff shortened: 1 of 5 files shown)


jayzhudev (approved):

jayzhudev:
Checked:

- Replacement order, ghost handling, and adaptation deltas match [CAR pseudocode](https://www.usenix.org/legacy/events/fast04/tech/full_papers/bansal/bansal.pdf).
- 20.8M differential operations matched an independent CAR model across capacities 0–12.

Test structure:

- Capacities: `0..=12` — 13 values.
- Seeds: `1..=64` — 64 deterministic streams.
- Operations per stream: `25,000`.
- Total: `13 × 64 × 25,000 = 20,800,000`.

Each stream ran PR’s `CarCache<u8, u64>` beside separate `VecDeque`-based CAR implementation directly following paper pseudocode:

- `T1`/`T2`: ordered entries with reference bits.
- `B1`/`B2`: ordered ghost keys.
- `p`: adaptive T1 target.
- Replacement: T1/T2 hand processing and ghost-directory rules.

Workload used 24 keys:

- ~25% `get`.
- ~75% `put`, including new keys and updates.

Compared after operations:

- `get` hit/miss and returned value.
- `put` evicted key and value.
- Live cache length.
- Adaptation parameter `p`.

This exercised cache hits, reference-bit clearing, clock rotation, T1→T2 promotion, T1/T2 eviction, B1/B2 hits, ghost LRU deletion, `p` increases/decreases, slot reuse behavior, and zero-capacity handling.

Limits:

- Randomized, not exhaustive.
- Model may share a misunderstood paper interpretation.
- Only `u8` keys and `u64` values.
- No allocation-failure, concurrency, or pathological `Hash`/`Eq` testing.
- Internal rings/indexes checked indirectly through observable behavior and panics.

bmc-http/src/cache.rs
```diff
@@ -74,35 +78,46 @@ struct GhostNode<K> {
...
-        // If we're at capacity, remove LRU first
-        let evicted_key = if self.free_slots.is_empty() {
-            self.remove_lru()
+    fn acquire_slot(&mut self) -> Option<usize> {
+        if let Some(slot) = self.free_slots.pop() {
+            return Some(slot);
+        }
+        if self.entries.len() < self.capacity {
+            self.entries.push(None);
+            Some(self.entries.len() - 1)
```
jayzhudev, line 109: nit: before I saw `self.entries.push(None)` I thought this may overflow.. could use `saturating_sub` to be carefree 😀

yoks: Reads a little bit cryptic, but i would leave it for now.

PR #144, opened 2026-07-13
feat: csdl-compiler, oem support changes

yoks:
### Compiler fixes

- **Keyword-safe identifiers.** New `generator::rust::ident::escaped` centralizes Rust-keyword escaping: raw form (`r#type`) where allowed, a trailing underscore for path keywords (`crate_`/`self_`/`super_`/`Self_`), and a leading underscore for names that aren't valid identifiers at all (a lone `_`, empty, or digit-leading munge). Every identifier the generator emits — property, action, module, type, and enum-variant names — now routes through it, so a keyword-valued schema name is escaped instead of panicking at codegen. Previously only property/action names were handled, and via ad-hoc match arms that panicked on any keyword they didn't special-case.

- **`Edm.Double` support.** Added the `Edm.Double → f64` mapping to the generated `edm` module. This is the only change to the standard corpus (used by NVIDIA port-metrics schemas). 

- **Update structs along sparse version chains.** A type's `Update` struct embeds its base type's `Update`, but a read-only base under a writable derivation never emitted one (versioned Redfish schemas skip versions, so `v1_6_0`'s base can be `v1_0_0`). Added a `ForcedUpdate` pre-pass in `RustGenerator::new` that walks each updating type's base chain and forces the ancestors to emit an `Update` struct — without pulling in the settings-object machinery, which stays gated on genuine updatability. `generates_update()` predicates on `ComplexType`/`EntityType` replace the previously-inlined condition.

- **OEM action placement and wire format.** Bound actions are now placed in their *defining* schema's module rather than the binding target's, so an OEM `Reset` bound to `Chassis.OemActions` lands in `nvidia_chassis::` instead of colliding with DMTF's `chassis::ChassisResetAction`. The serde `rename` now uses the defining schema's root namespace (`#NvidiaChassis.Reset`), matching what Redfish actually serializes — a correctness fix for the previous binding-parameter-based key. Added `Namespace::root()`/`truncated()`.

- **Deterministic output.** A module shared across schema versions recorded whichever type's namespace happened to be inserted first (HashMap order), leaking a versioned namespace into a doc comment nondeterministically. Modules now record only the namespace prefix that names them (`truncated(depth + 1)`), so generation is byte-stable across runs — a prerequisite for CI diff-verification.

- **Setter naming.** Digit-leading properties (`128BitCasCompleterSupported` → field `_128_bit_...`) produced `with__128_bit_...` (double underscore, non-snake-case warning); the setter now trims the leading underscore to `with_128_bit_...`.


poroh (changes requested):
Could you please split this PR to individual changes?

PR #162, opened 2026-07-16
fix(accounts): Prevent stale slot reuse during account creation

### LINKED ISSUE #141: Slot-defined account creation reuses stale snapshot, can silently clobber accounts (TOCTOU)
## Summary

**Severity:** LOW
**Category:** `toctou_stale_state`
**File:** `redfish/src/account/collection.rs:166`

## Description

For slot-based BMCs, `create_account` (`collection.rs:166-213`) scans a snapshot of members captured at collection construction (lines 123-140). Expanded members are never re-fetched (`NavProperty::get` returns an `Arc` clone of the snapshot, `core/src/nav_property.rs:219-224`).

When the chosen slot resource carries no ETag, the `PATCH` falls back to `If-Match: *` (`bmc-http/src/lib.rs:712-714`), which matches any state. Two concurrent creators deterministically select the same free slot; the second `PATCH` silently overwrites the first's credentials and both callers report success.

## Impact

Silent account credential overwrite under concurrent provisioning. The first operator's automation is left unable to authenticate with no error reported.

## Exploit Scenario

Two operators or orchestration jobs provision accounts concurrently; both pick the same slot; the second `PATCH` replaces the first's username/password without any conflict error.

## Recommendation

- Re-fetch the candidate slot immediately before the enabling `PATCH` and require `Enabled == false` on the fresh read.
- Never fall back to `If-Match: *` for this operation.
- Verify `UserName` matches expectations after the `PATCH`.

---
*Reported by automated scan — 2026-07-11*

nvlitagaki:
## What changed

- Re-fetch a slot-defined account immediately before enabling it instead of relying on the expanded collection snapshot.
- Require the fresh account to remain disabled and carry an ETag before issuing the update, so the HTTP path cannot fall back to `If-Match: *` for slot creation.
- Add regressions for a candidate that became enabled after collection expansion and for a disabled candidate without an ETag.

## Why

Concurrent creators could select the same disabled slot from the collection snapshot. If the snapshot did not carry an ETag, the second update became unconditional and could overwrite the first account's credentials.

The fresh read plus conditional update closes that race: a slot already claimed is skipped, while concurrent updates against the same fresh ETag let the BMC reject the stale writer.

## Validation

- Before the production change, `create_account_slot_defined_rechecks_stale_candidate` failed because the code issued an update for slot 3 before the expected fresh GET.
- `cargo test -p nv-redfish-tests --test tests-account-service` (15 passed)
- `cargo test -p nv-redfish-tests --tests` (all integration and trybuild targets passed)
- `cargo test -p nv-redfish --features accounts`
- `cargo test -p nv-redfish-bmc-http --test reqwest_client_tests --features reqwest test_patch_update_request`
- `cargo fmt --all -- --check`
- `cargo clippy -p nv-redfish --features accounts -- -D warnings`
- `cargo clippy -p nv-redfish-tests --test tests-account-service -- -D warnings`

The checks ran with Rust 1.90.0 in a Linux arm64 container. A physical Dell BMC was not available, so this remains a draft pending maintainer review or hardware-backed verification.

Resolves #141

### DIFF
--- a/redfish/src/account/collection.rs
+++ b/redfish/src/account/collection.rs
@@ -185,6 +185,21 @@ impl<B: Bmc> AccountCollection<B> {
                     // Slot is already explicitly enabled. Find another slot.
                     continue;
                 }
+
+                // Expanded collection members are a snapshot. Re-fetch the
+                // candidate immediately before updating it so concurrent
+                // account creation cannot reuse stale slot state.
+                let fresh_nav = NavProperty::new_reference(account.raw().odata_id().clone());
+                let Ok(account) = Account::new(&self.bmc, &fresh_nav, &self.config.account).await
+                else {
+                    continue;
+                };
+                if account.is_enabled() || account.raw().etag().is_none() {
+                    // Without a fresh ETag, the HTTP BMC would fall back to an
+                    // unconditional `If-Match: *` update.
+                    continue;
+                }
+
                 // Build an update based on the create request:
                 let update = ManagerAccountUpdate {
                     base: None,
--- a/redfish/src/account/collection.rs
+++ b/redfish/src/account/collection.rs
@@ -188,17 +188,18 @@ impl<B: Bmc> AccountCollection<B> {
 
                 // Expanded collection members are a snapshot. Re-fetch the
                 // candidate immediately before updating it so concurrent
-                // account creation cannot reuse stale slot state.
-                let fresh_nav = NavProperty::new_reference(account.raw().odata_id().clone());
-                let Ok(account) = Account::new(&self.bmc, &fresh_nav, &self.config.account).await
-                else {
+                // account creation cannot reuse stale slot state. Require a
+                // fresh ETag so the HTTP BMC cannot fall back to `If-Match: *`.
+                let Some(account) = Account::new(
+                    &self.bmc,
+                    &NavProperty::new_reference(account.raw().odata_id().clone()),
+                    &self.config.account,
+                )
+                .await
+                .ok()
+                .filter(|a| !a.is_enabled() && a.raw().etag().is_some()) else {
                     continue;
                 };
-                if account.is_enabled() || account.raw().etag().is_none() {
-                    // Without a fresh ETag, the HTTP BMC would fall back to an
-                    // unconditional `If-Match: *` update.
-                    continue;
-                }
 
                 // Build an update based on the create request:
                 let update = ManagerAccountUpdate {
--- a/tests/tests/tests-account-service.rs
+++ b/tests/tests/tests-account-service.rs
@@ -243,8 +243,14 @@ async fn get_account_collection(
     Ok(account_service.accounts().await.map(Option::unwrap)?)
 }
 
-fn slot_member(accounts_id: &str, id: u32, enabled: bool, user_name: &str) -> JsonValue {
-    json!({
+fn slot_member(
+    accounts_id: &str,
+    id: u32,
+    enabled: bool,
+    user_name: &str,
+    etag: Option<&str>,
+) -> JsonValue {
+    let mut member = json!({
         ODATA_ID: format!("{accounts_id}/{id}"),
         ODATA_TYPE: MANAGER_ACCOUNT_DATA_TYPE,
         "Id": id.to_string(),
@@ -252,20 +258,11 @@ fn slot_member(accounts_id: &str, id: u32, enabled: bool, user_name: &str) -> Js
         "Enabled": enabled,
         "AccountTypes": [],
         "UserName": user_name,
-    })
-}
-
-fn slot_member_with_etag(
-    accounts_id: &str,
-    id: u32,
-    enabled: bool,
-    user_name: &str,
-    etag: &str,
-) -> JsonValue {
-    json_merge([
-        &slot_member(accounts_id, id, enabled, user_name),
-        &json!({ "@odata.etag": etag }),
-    ])
+    });
+    if let Some(etag) = etag {
+        member["@odata.etag"] = JsonValue::String(etag.to_string());
+    }
+    member
 }
 
 async fn account_fixture(
@@ -280,7 +277,9 @@ async fn account_fixture(
     let members = JsonValue::Array(
         slots
             .iter()
-            .map(|&(id, enabled, user_name)| slot_member(&accounts_id, id, enabled, user_name))
+            .map(|&(id, enabled, user_name)| {
+                slot_member(&accounts_id, id, enabled, user_name, None)
+            })
             .collect(),
     );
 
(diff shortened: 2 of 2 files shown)


redfish/src/account/collection.rs
```diff
@@ -185,6 +185,21 @@ impl<B: Bmc> AccountCollection<B> {
...
+                let fresh_nav = NavProperty::new_reference(account.raw().odata_id().clone());
+                let Ok(account) = Account::new(&self.bmc, &fresh_nav, &self.config.account).await
+                else {
+                    continue;
+                };
+                if account.is_enabled() || account.raw().etag().is_none() {
+                    // Without a fresh ETag, the HTTP BMC would fall back to an
+                    // unconditional `If-Match: *` update.
+                    continue;
+                }
```
yoks:
```suggestion
let Some(account) = Account::new(
    &self.bmc,
    &NavProperty::new_reference(account.raw().odata_id().clone()),
    &self.config.account,
)
.await
.ok()
.filter(|a| !a.is_enabled() && a.raw().etag().is_some())
else {
    continue;
};
```
I would do it in more ideomatic way here.

tests/tests/tests-account-service.rs
```diff
@@ -255,6 +255,19 @@ fn slot_member(accounts_id: &str, id: u32, enabled: bool, user_name: &str) -> Js
     })
 }
 
+fn slot_member_with_etag(
```
yoks: I would extend already existing `slot_member` function with Optional etag

yoks:
@nvlitagaki PR open as a draft, do you expect to work on it or is it finished?

nvlitagaki:
This is staying in draft until I can get real hardware to verify on

yoks:
@nvlitagaki any updates on this PR?

nvlitagaki:
> @nvlitagaki any updates on this PR?

@yoks I finally got my hands on real Dell hardware to test, and unfortunately this fix doesn't work as written. What actually happened was that both concurrent updates were accepted by the iDRAC, even with the ETag change. I'll be working on an alternate fix.

nvlitagaki:
Testing this fix revealed a concurrency bug in iDRAC implementation of conditional Redfish updates: the If-Match is enforced for sequential requests but not atomically for concurrent PATCHes.

### How this bug was tested:

#### Test setup
Tests were run by making `curl` requests from an environment with direct network access to the iDRAC, deliberately bypassing nv-redfish logic to isolate BMC behavior.

BMC administrator credentials were retrieved through the existing secret-management path and placed in a temporary mode-0600 netrc file

Before starting, did the following:
- Confirmed that the Redfish service identified itself as Dell iDRAC.
- Enumerated all account slots.
- Selected an empty, disabled, non-administrator slot.
- Recorded the complete observable state of that slot.
- Temporarily disabled System Lockdown so that account modification testing could proceed

#### Testing

#####  set up JSON payload
```
USERNAME_A='etag-test-a'
USERNAME_B='etag-test-b'

PASSWORD_A='REDACTED_TEMP_PASSWORD_A'
PASSWORD_B='REDACTED_TEMP_PASSWORD_B'

jq -n \
  --arg username "${USERNAME_A}" \
  --arg password "${PASSWORD_A}" \
  '{
    UserName: $username,
    Password: $password,
    RoleId: "Operator",
    Enabled: true
  }' > "${WORK_DIR}/patch-a.json"

jq -n \
  --arg username "${USERNAME_B}" \
  --arg password "${PASSWORD_B}" \
  '{
    UserName: $username,
    Password: $password,
    RoleId: "Operator",
    Enabled: true
  }' > "${WORK_DIR}/patch-b.json"

chmod 600 \
  "${WORK_DIR}/patch-a.json" \
  "${WORK_DIR}/patch-b.json"
```

#####  made concurrent requests with same ETag
```
rm -f "${WORK_DIR}/start"

(
  while [[ ! -e "${WORK_DIR}/start" ]]; do
    :
  done

  curl --http1.1 \
    -ksS \
    --max-time 60 \
    --netrc-file "${NETRC}" \
    -X PATCH \
    -H 'Content-Type: application/json' \
    -H "If-Match: ${ETAG_A}" \
    --data-binary "@${WORK_DIR}/patch-a.json" \
    -D "${WORK_DIR}/patch-a.headers" \
    -o "${WORK_DIR}/patch-a.response" \
    -w '%{http_code} %{time_total}\n' \
    "${BASE_URL}${ACCOUNT_URI}" \
    > "${WORK_DIR}/patch-a.status"
) &
PID_A=$!

(
  while [[ ! -e "${WORK_DIR}/start" ]]; do
    :
  done

  curl --http1.1 \
    -ksS \
    --max-time 60 \
    --netrc-file "${NETRC}" \
    -X PATCH \
    -H 'Content-Type: application/json' \
    -H "If-Match: ${ETAG_B}" \
    --data-binary "@${WORK_DIR}/patch-b.json" \
    -D "${WORK_DIR}/patch-b.headers" \
    -o "${WORK_DIR}/patch-b.response" \
    -w '%{http_code} %{time_total}\n' \
    "${BASE_URL}${ACCOUNT_URI}" \
    > "${WORK_DIR}/patch-b.status"
) &
PID_B=$!

touch "${WORK_DIR}/start"

wait "${PID_A}"
wait "${PID_B}"
```

##### checked output and inspected result slot
```
Client A: HTTP 200
Client B: HTTP 200
```

queried what got saved to slot
```
curl --http1.1 \
  -ksS \
  --netrc-file "${NETRC}" \
  -D "${WORK_DIR}/final.headers" \
  -o "${WORK_DIR}/final.json" \
  "${BASE_URL}${ACCOUNT_URI}"

jq '{
  UserName,
  RoleId,
  Enabled,
  Locked,
  etag: ."@odata.etag"
}' "${WORK_DIR}/final.json"
```
and got:

```
{
  "UserName": "etag-test-a",
  "RoleId": "Operator",
  "Enabled": true,
  "Locked": false,
  "etag": "<etag-after-race>"
}
```

nvlitagaki:
The question now is: Do we file a bug with Dell, or do we try to do a workaround to account for the iDRAC bug.

yoks:
This is clearly are firmware bug, a it violates voth RFC 9110 - 13.1.1 and Redfish spec. So i would still add fix which uses etags as it is propper up to spec way (this PR) and will file bug with Dell for this.

nvlitagaki:
> This is clearly are firmware bug, a it violates voth RFC 9110 - 13.1.1 and Redfish spec. So i would still add fix which uses etags as it is propper up to spec way (this PR) and will file bug with Dell for this.

@yoks Sounds good. I moved this to ready for review -- at least it will address the case of two requests in very quick succession.

yoks (approved):

PR #165, opened 2026-07-20
fix(quirks): enable missing-ChassisType/Name workaround for Supermicro

hatamzad-nv:
## Summary
Supermicro host BMCs expose chassis (e.g. a `SmartNIC` chassis) that omit the Redfish-**Required** `ChassisType` (and `Name`) fields, which fails chassis deserialization in consumers. This classifies `Vendor=Supermicro` as a platform and enables the existing `bug_missing_chassis_type_field` / `bug_missing_chassis_name_field` workarounds for it — the patch injects a default only when the field is absent (`entry().or_insert()`), so it's a no-op for compliant Supermicro chassis.

Reported downstream in NVIDIA/infra-controller#3715.

## Changes
- `bmc_quirks.rs`: add `Platform::Supermicro` (detected via service-root `Vendor == "Supermicro"`) and enable the two chassis workarounds for it.
- Tests: `supermicro_missing_chassis_type_workaround` and `supermicro_missing_chassis_name_workaround`, plus a `supermicro_service_root` helper.

## Notes for review
- I enabled the `Name` workaround alongside `ChassisType`: the reported `SmartNIC` chassis omitted every field, and `Name` is likewise Required, so ChassisType alone would just move the error to `missing field Name`. Mirrors the AmiViking precedent. Happy to drop it to ChassisType-only if preferred.
- Detection is `Vendor == "Supermicro"` (all Supermicro BMCs). Can narrow to a specific product/version if that's preferred — but the patch is harmless for compliant chassis.

## Testing
- `cargo test -p nv-redfish-tests` — all green (17/17 chassis tests).
- `cargo clippy -p nv-redfish --features chassis --all-targets` — clean.

### DIFF
--- a/redfish/src/bmc_quirks.rs
+++ b/redfish/src/bmc_quirks.rs
@@ -37,6 +37,7 @@ enum Platform {
     NvidiaDpu,
     Anonymous1_9_0,
     NvSwitch,
+    Supermicro,
 }
 
 impl BmcQuirks {
@@ -63,6 +64,7 @@ impl BmcQuirks {
             Some("NVIDIA") if product_str == Some("P3809") => Some(Platform::NvSwitch),
             Some("NVIDIA") => Some(Platform::Nvidia),
             Some("Nvidia") if product_str == Some("Nvidia-BMCMezz") => Some(Platform::NvidiaDpu),
+            Some("Supermicro") => Some(Platform::Supermicro),
             None if redfish_version_str == Some("1.9.0") => Some(Platform::Anonymous1_9_0),
             _ => None,
         };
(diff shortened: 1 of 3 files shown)


hatamzad-nv:
@poroh this is the nv-redfish quirk you asked for on NVIDIA/infra-controller#3726 (now closed in favor of this). Mind taking a look? Two things I'd value your call on, noted above: enabling the `Name` workaround alongside `ChassisType`, and the `Vendor == "Supermicro"` detection (I can narrow it if you'd rather gate on product/version).

poroh (changes requested):
I would prefer to narrow down to specific platform because most SMC servers I've seen provide ChassisType. We should be able to identify this platform in future and remove this patch if it will be fixed at some point of time by BMC firmware update. Thanks.

krish-nvidia:
Poked around on the actual BMC and found that the `/redfish/v1/Chassis/SmartNIC_1` returns the following:
```
{
    "error": {
        "code": "Base.1.10.GeneralError",
        "message": "A general error has occurred. See ExtendedInfo for more information.",
        "@Message.ExtendedInfo": [
            {
                "MessageId": "SMC.1.0.OemLicenseNotPassed",
                "Severity": "Warning",
                "Resolution": "Please activate at least one license in above message.",
                "Message": "Not licensed to perform this request. The following licenses DCMS  were needed",
                "MessageArgs": [
                    "DCMS "
                ],
                "RelatedProperties": [
                    ""
                ]
            }
        ]
    }
}
```
The other two chassis's in the response return `ChassisType` as expected. This fix needs to handle the error above, not just a `bug_missing_chassis_type_field`.

david-mateer:
Thanks Krish, I didn't know about this so I've asked Sean how this was done originally.  The system board replacement has maybe invalidated the license.
I guess the question now is, would anyone deploy a Supermicro with no SmartNIC_1 license?  If not, maybe this is not a bug?  I don't know the answer though.

Best regards,

David.

hatamzad-nv:
Closing, this turned out to be a SmartNIC license issue after a board swap, not a BMC quirk. Nothing to fix here. Thanks all.

PR #166, opened 2026-07-20
fix(redfish): tolerate null collection Members on BlueField BMC

shevchenko-evgeny:
This is a per-collection opt-in: collections not routed through CollectionWithPatch (e.g. processors, memory) are unaffected until explicitly migrated.

This is another option for addressing the problem discussed more generally in [PR #120](https://github.com/NVIDIA/nv-redfish/pull/120).

I’m still not sure whether it is better than the solution proposed there. That approach only covers collections explicitly gated through `CollectionWithPatch`, while it appears that any empty collection may exhibit the same issue on that BMC. As a result, we could still see unexpected behavior in certain states.

At the same time, routing every possible collection through `CollectionWithPatch` does not seem practical and would be quite verbose.


tests/tests/test-computer-system.rs
```diff
@@ -238,6 +181,134 @@ async fn nvidia_dpu_empty_system_uuid_in_expanded_members_workaround(
...
+        &ids.systems_id,
+        json!({
+            ODATA_ID: &ids.systems_id,
+            ODATA_TYPE: &SYSTEM_COLLECTION_DATA_TYPE,
+            "Id": resource_name(&ids.systems_id),
+            "Name": "Computer System Collection",
+            "Members": [
+                computer_system(&ids, json!({
+                "Storage": { "@odata.id": "/redfish/v1/Systems/Bluefield/Storage" },
+                "Boot": {
```
poroh, line 258: Could you please minimize JSON noise from the test. I mean leave here only relevant parts of the original jsons that are checked by the test. Otherwise, it is harder to follow what this test is checking.

redfish/src/computer_system/mod.rs
```diff
@@ -198,3 +202,54 @@ fn normalize_empty_uuid_field(mut v: JsonValue) -> JsonValue {
     }
     v
 }
+
+/// Storage collection.
```
poroh: This change looks not relevant to the PR description.

poroh (changes requested):

redfish/src/computer_system/mod.rs
shevchenko-evgeny: While fixing this I realized the `Members: null` problem isn't specific to `BootOptions` - every empty collection on the BlueField BMC serializes Members as null instead of []. So far I've found at least these:

`/redfish/v1/Cables`
`/redfish/v1/AccountService/MultiFactorAuth/ClientCertificate/Certificates`
`/redfish/v1/Systems/Bluefield/Oem/Nvidia/Truststore/Certificates`
`/redfish/v1/Systems/Bluefield/Storage`
`/redfish/v1/Systems/Bluefield/BootOptions`

First three (Cables, both Certificates) - not an big issue today: their collection types aren't compiled into the tree, so they're only ever held as references  and never deserialized. The null is harmless until something actually fetches them.
Latest two I fixed explicitly.

tests/tests/test-computer-system.rs
shevchenko-evgeny: Done

redfish/src/computer_system/mod.rs
poroh: I've never seen something like this from BF3 DPU. Maybe you are using some old BMC FW version?

shevchenko-evgeny: It’s definitely not the latest version, but until NICO upgrades the BMC firmware, it should be sufficient to pass the pre-injection phase.

For context, this is part of the [PowerEdge XE9780 onboarding](https://github.com/NVIDIA/infra-controller/issues/3227) umbrella issue.

poroh: Ok. Fair enough. Could you please move storage collection to storage.rs module then. Everything else LGTM.

shevchenko-evgeny: Done

poroh (dismissed):

poroh:
@shevchenko-evgeny could you please run `cargo fmt` and update your commit?

shevchenko-evgeny:
> @shevchenko-evgeny could you please run `cargo fmt` and update your commit?

Done

yoks:
@shevchenko-evgeny CI still failing, can you try run `make ci` locally to ensure it compiles at leat on your machine?

Thanks.

shevchenko-evgeny:
@yoks @poroh clippy failed due to `.map(|r| Some(r))` - fixed now. Hope it helps.

yoks (approved):

PR #167, opened 2026-07-21
fix: Adding a event size cap and stream timeout to prevent compromised BMC…

### LINKED ISSUE #138: SSE stream buffers malicious BMC data unbounded with timeout disabled
## Summary

**Severity:** MEDIUM
**Category:** `uncontrolled_resource_consumption`
**File:** `bmc-http/src/reqwest.rs:1073`

## Description

`Client::sse` sets `.timeout(Duration::MAX)` then feeds `response.bytes_stream()` into `sse_stream::SseStream::from_byte_stream` (`reqwest.rs:1085`) with no cap on buffered event size. SSE framing buffers a `data:` line until a terminator arrives; a server that never sends one grows the buffer indefinitely. The usual total-request timeout backstop is deliberately removed on this path.

Reachable via `EventService::events()` → `HttpBmc::stream` (`bmc-http/src/lib.rs:825-838`) → `Client::sse`.

## Impact

A compromised BMC can exhaust the client process's memory, killing the daemon and taking down monitoring of every other BMC managed by the same process.

## Exploit Scenario

Client subscribes to BMC events; compromised BMC responds with a valid SSE header then streams gigabytes of `data:` payload with no newline; client memory grows until the OS kills the process.

## Recommendation

- Enforce a maximum unterminated-event size (e.g. 1 MiB) and abort the stream with an error when exceeded.
- Expose an SSE idle/read timeout in `ClientParams` instead of hardcoding `Duration::MAX`.

---
*Reported by automated scan — 2026-07-11*

rohanharkin:
#138 
Adds a configurable per-event byte cap that aborts the SSE stream if a single event exceeds the limit without terminating, preventing a compromised BMC from exhausting memory by sending an endless unterminated event.
Adds a configurable idle timeout that aborts the stream if no event arrives within the window, catching connections that have stalled

This was tested manually using a mock BMC that opens a text/event-stream response and then streams an SSE event that never terminates, against a client that consumes the stream through EventService/Bmc::stream in a memory-capped container.

Two tests added as well
sse_aborts_when_event_exceeds_byte_limit()
sse_aborts_on_idle_timeout()

### DIFF
--- a/bmc-http/src/reqwest.rs
+++ b/bmc-http/src/reqwest.rs
@@ -558,7 +558,7 @@ impl ClientParams {
 
     /// Sets the maximum buffered size of a single, not-yet-terminated SSE event.
     ///
-    /// See [`ClientParams::sse_max_event_bytes`].
+    /// See [`SseOptions::max_event_bytes`].
     #[must_use]
     pub const fn sse_max_event_bytes(mut self, bytes: usize) -> Self {
         self.sse.max_event_bytes = bytes;
@@ -581,9 +581,8 @@ impl ClientParams {
 /// reqwest HTTP client library. It supports all standard HTTP features including
 /// TLS, authentication, and connection pooling.
 #[derive(Clone)]
-#[allow(clippy::struct_field_names)]
 pub struct Client {
-    client: ReqwestClient,
+    inner: ReqwestClient,
     retry: Option<RetryPolicy>,
     sse: SseOptions,
 }
@@ -655,7 +654,7 @@ impl Client {
         }
 
         Ok(Self {
-            client: builder.build()?,
+            inner: builder.build()?,
             retry: params.retry,
             sse: params.sse,
         })
@@ -671,14 +670,11 @@ impl Client {
     /// The supplied client must reject cross-origin redirects. Reqwest's default redirect policy
     /// can forward Redfish `X-Auth-Token` and arbitrary custom headers to another origin.
     #[must_use]
-    pub const fn with_client(client: ReqwestClient) -> Self {
+    pub fn with_client(client: ReqwestClient) -> Self {
         Self {
-            client,
+            inner: client,
             retry: None,
-            sse: SseOptions {
-                max_event_bytes: 1024 * 1024, // 1 MiB
-                idle_timeout: None,
-            },
+            sse: SseOptions::default(),
         }
     }
 }
@@ -690,7 +686,7 @@ impl Client {
     /// bodies cannot be cloned and are sent exactly once.
     async fn send(&self, request: reqwest::Request) -> Result<reqwest::Response, BmcError> {
         let Some(policy) = &self.retry else {
-            return Ok(self.client.execute(request).await?);
+            return Ok(self.inner.execute(request).await?);
         };
 
         let mut attempt: u32 = 0;
@@ -700,7 +696,7 @@ impl Client {
             // try_clone() returns None for streaming bodies, which therefore
             // get a single attempt.
             let next = if is_last { None } else { current.try_clone() };
-            let response = self.client.execute(current).await?;
+            let response = self.inner.execute(current).await?;
             match next {
                 // The clone is identical to the request just sent, so the
                 // classifier sees what went over the wire.
@@ -1078,7 +1074,7 @@ impl HttpClient for Client {
         T: DeserializeOwned,
     {
         let mut request =
-            auth_headers(self.client.get(url), credentials).headers(custom_headers.clone());
+            auth_headers(self.inner.get(url), credentials).headers(custom_headers.clone());
 
         if let Some(etag) = etag {
             request = request.header(header::IF_NONE_MATCH, etag.to_string());
@@ -1099,7 +1095,7 @@ impl HttpClient for Client {
         B: Serialize + Send + Sync,
         T: DeserializeOwned + Send + Sync,
     {
-        let request = auth_headers(self.client.post(url), credentials)
+        let request = auth_headers(self.inner.post(url), credentials)
             .headers(custom_headers.clone())
             .json(body);
 
@@ -1118,7 +1114,7 @@ impl HttpClient for Client {
         T: DeserializeOwned + Send + Sync,
     {
         let request = self
-            .client
+            .inner
             .post(url)
             .headers(custom_headers.clone())
             .json(body);
@@ -1140,7 +1136,7 @@ impl HttpClient for Client {
         T: DeserializeOwned + Send + Sync,
     {
         let mut request =
-            auth_headers(self.client.patch(url), credentials).headers(custom_headers.clone());
+            auth_headers(self.inner.patch(url), credentials).headers(custom_headers.clone());
 
         request = request.header(header::IF_MATCH, etag.to_string());
 
@@ -1158,7 +1154,7 @@ impl HttpClient for Client {
         T: DeserializeOwned + Send + Sync,
     {
         let request =
-            auth_headers(self.client.delete(url), credentials).headers(custom_headers.clone());
+            auth_headers(self.inner.delete(url), credentials).headers(custom_headers.clone());
 
         let response = self.send(request.build()?).await?;
         self.handle_modification_response(response).await
@@ -1205,7 +1201,7 @@ impl HttpClient for Client {
             form = form.part(name, part);
         }
 
-        let request = auth_headers(self.client.post(url), credentials)
+        let request = auth_headers(self.inner.post(url), credentials)
             .headers(custom_headers.clone())
             .multipart(form)
             .timeout(upload_timeout);
@@ -1238,7 +1234,7 @@ impl HttpClient for Client {
 
         let body = reqwest::Body::wrap_stream(ReaderStream::new(reader.compat()));
 
-        let mut request = auth_headers(self.client.post(url), credentials)
+        let mut request = auth_headers(self.inner.post(url), credentials)
             .headers(custom_headers.clone())
             .header(header::CONTENT_TYPE, "application/octet-stream")
             .body(body)
@@ -1262,7 +1258,7 @@ impl HttpClient for Client {
         credentials: &BmcCredentials,
         custom_headers: &HeaderMap,
     ) -> Result<BoxTryStream<T, Self::Error>, Self::Error> {
-        let request = auth_headers(self.client.get(url), credentials)
+        let request = auth_headers(self.inner.get(url), credentials)
             .headers(custom_headers.clone())
             .header(header::ACCEPT, "text/event-stream")
             .timeout(Duration::MAX);
--- a/bmc-http/src/reqwest.rs
+++ b/bmc-http/src/reqwest.rs
@@ -17,8 +17,7 @@
 
 use std::error::Error as StdErr;
 use std::fmt;
-use std::sync::atomic::AtomicUsize;
-use std::sync::atomic::Ordering;
+use std::future::ready;
 use std::sync::Arc;
 use std::time::Duration;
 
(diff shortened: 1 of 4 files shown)


bmc-http/src/reqwest.rs
```diff
@@ -286,6 +356,16 @@ pub struct ClientParams {
...
+    /// Maximum bytes buffered for a single, not-yet-terminated SSE event before
+    /// [`Client::sse`] aborts with [`BmcError::SseEventTooLarge`]. Guards against
+    /// a server that never sends the SSE event terminator.
+    pub sse_max_event_bytes: usize,
+    /// Maximum idle time between SSE events before [`Client::sse`] aborts with
+    /// [`BmcError::SseIdleTimeout`]. `None` disables the check.
+    pub sse_idle_timeout: Option<Duration>,
+    /// Optional overall deadline applied to the SSE request. `None` leaves the
+    /// long-lived SSE request without a total-request timeout.
+    pub sse_total_timeout: Option<Duration>,
```
yoks: I do not like setting this one as None by default, should be MAX.

Reason is:

1. It is how it behaves right now, will break existing clients
2. SSE connection are long lived and as long as there is data it should live, forcing user to set it is not as friendly
3. You already have sse_idle timeout, why do we want total timeout here?

bmc-http/src/reqwest.rs
```diff
@@ -302,6 +382,9 @@ impl Default for ClientParams {
             default_headers: None,
             use_rust_tls: true,
             retry: None,
+            sse_max_event_bytes: DEFAULT_SSE_MAX_EVENT_BYTES,
```
yoks: If we only use it Default, it feels what define the const is redundant. 1mb can be defined here, as it consistent with other default values

bmc-http/src/reqwest.rs
```diff
@@ -1082,21 +1217,81 @@ impl HttpClient for Client {
...
+                }
+                let next = match idle {
+                    Some(d) => match timeout(d, events.next()).await {
+                        Ok(item) => item,
+                        Err(_) => Some(Err(BmcError::SseIdleTimeout { idle: d })),
+                    },
+                    None => events.next().await,
+                };
+                match next {
+                    Some(item @ Err(_)) => Some((item, (events, true))),
```
yoks: Any error will terminate the stream here, this should not be true for JSON errors, as they are event based.

yoks:
SSEStream implementation discrards comments, this means for healthy stream which sends heartbeats via comments would not:

- Count towards byte-limit (same issue with unbounded growth)
- Would terminate idle stream even if it sends heartbeats

I do not know how to properly fix it with SSEStream as it not supports it. So either commnet and contribution to upstream lib to fix it. Or rewrite SSE stream ourselves to properly fix it.

bmc-http/src/reqwest.rs
```diff
@@ -190,6 +259,7 @@ type RetryClassifier =
 /// let params = ClientParams::new().retry(policy);
 /// ```
 #[derive(Clone)]
+#[allow(clippy::struct_field_names)]
```
yoks: It does seems like unrelated change.

bmc-http/src/reqwest.rs
```diff
@@ -389,6 +472,40 @@ impl ClientParams {
...
+    /// See [`ClientParams::sse_idle_timeout`].
+    #[must_use]
+    pub const fn sse_idle_timeout(mut self, timeout: Duration) -> Self {
+        self.sse_idle_timeout = Some(timeout);
+        self
+    }
+
+    /// Disables the SSE idle timeout.
+    #[must_use]
+    pub const fn no_sse_idle_timeout(mut self) -> Self {
```
yoks: None is the default value, so this looks like a bloat (this function does not seems to be useful)

bmc-http/src/reqwest.rs
rohanharkin: before adding this CI was failing as the additional fields went past Clippys struct_field_name_threshold. It was then failing because of the field client within Client

yoks: Well, this is valid lint then. As your sse prefixed params can be moved to SSE options struct

yoks:
After playing with Claude with this change, it looks like this would be a better shape here:

```rust
#[derive(Clone, Copy)]
enum DelimiterState {
    /// Inside a line.
    Data,
    /// Saw `\r`; a following `\n` belongs to the same terminator.
    Cr,
    /// Saw one complete line terminator.
    Eol,
}

// This should fix the issue with heartbeat and byte counting, as we look at raw data
fn cap_event_bytes(
    bytes: impl Stream<Item = Result<Bytes, reqwest::Error>>,
    limit: usize,
) -> impl Stream<Item = Result<Bytes, SseByteError>> {
    use DelimiterState::{Cr, Data, Eol};
    bytes.scan((0_usize, Data), move |(count, state), chunk| {
        let item = chunk.map_err(SseByteError::Upstream).and_then(|bytes| {
            for &b in bytes.iter() {
                *count += 1;
                *state = match (*state, b) {
                    (Data, b'\r') => Cr,
                    (Data | Cr, b'\n') => Eol,
                    (Cr | Eol, b'\r') => {
                        *count = 0; // blank line: event delimited
                        Cr
                    }
                    (Eol, b'\n') => {
                        *count = 0; // blank line: event delimited
                        Eol
                    }
                    _ => Data,
                };
            }
            if *count > limit {
                Err(SseByteError::EventTooLarge { limit })
            } else {
                Ok(bytes)
            }
        });
        future::ready(Some(item))
    })
}

fn event_to_item<T: DeserializeOwned>(
    event: Result<sse_stream::Sse, sse_stream::Error>,
) -> Option<Result<T, BmcError>> {
    match event {
        Err(err) => Some(Err(map_sse_error(err))),
        Ok(sse) => sse.data.map(|data| {
            serde_path_to_error::deserialize(&mut serde_json::Deserializer::from_str(&data))
                .map_err(BmcError::JsonError)
        }),
    }
}

fn with_idle_timeout<T>(
    events: impl Stream<Item = Result<T, BmcError>>,
    idle: Option<Duration>,
) -> impl Stream<Item = Result<T, BmcError>> {
    match idle {
        // tokio_stream is additional dependency, but it should be fine 
        Some(d) => tokio_stream::StreamExt::timeout(events, d)
            .map(move |item| item.unwrap_or(Err(BmcError::SseIdleTimeout { idle: d })))
            .left_stream(),
        None => events.right_stream(),
    }
}

fn fuse_on_fatal<T>(
    events: impl Stream<Item = Result<T, BmcError>>,
) -> impl Stream<Item = Result<T, BmcError>> {
    events.scan(false, |errored, item| {
        // is_stream_fatal is impl on BmcError which checks if it JSON error or not, so we would
        // terminate on non fata errors
        let stop = std::mem::replace(errored, matches!(&item, Err(e) if e.is_stream_fatal()));
        future::ready((!stop).then_some(item))
    })
}
```

Ans SSE function can be compacted to this:
```rust
async fn sse<T: Send + Sized + for<'de> serde::Deserialize<'de>>(
    &self,
    url: Url,
    credentials: &BmcCredentials,
    custom_headers: &HeaderMap,
) -> Result<BoxTryStream<T, Self::Error>, Self::Error> {
    let request = auth_headers(self.client.get(url), credentials)
        .headers(custom_headers.clone())
        .header(header::ACCEPT, "text/event-stream")
        .timeout(self.sse_total_timeout.unwrap_or(Duration::MAX)); // note what it does default to MAX here

    let response = self.send(request.build()?).await?;

    if response.status() != StatusCode::OK {
        return Err(BmcError::HttpError {
            status: response.status(),
            error_response: None,
        });
    }

    let capped = cap_event_bytes(response.bytes_stream(), self.sse_max_event_bytes);
    let events = sse_stream::SseStream::from_bytes_stream(capped)
        .filter_map(|event| future::ready(event_to_item(event)));

    Ok(Box::pin(fuse_on_fatal(with_idle_timeout(
        events,
        self.sse_idle_timeout,
    ))))
}

```

bmc-http/src/reqwest.rs
rohanharkin: Sorry I was confused - I added this in the wrong place

bmc-http/src/reqwest.rs
rohanharkin: Yeah this makes sense - total timeout is not needed

bmc-http/src/reqwest.rs
rohanharkin: Missed that - thanks

rohanharkin:
> SSEStream implementation discrards comments, this means for healthy stream which sends heartbeats via comments would not:
> 
> * Count towards byte-limit (same issue with unbounded growth)
> * Would terminate idle stream even if it sends heartbeats
> 
> I do not know how to properly fix it with SSEStream as it not supports it. So either commnet and contribution to upstream lib to fix it. Or rewrite SSE stream ourselves to properly fix it.

It seems like a lot to rewrite SSE for this change - Probably makes more sense as you suggested to comment it and contribute to SSE

bmc-http/src/reqwest.rs
```diff
@@ -1082,21 +1264,42 @@ impl HttpClient for Client {
...
+            .filter_map(|event| async move { event_to_item::<T>(event) });
+
+        // Liveness bound: optionally abort if no event arrives within the idle
+        // window. It is keyed on decoded events, so comment heartbeats (which
+        // `sse_stream` discards) do NOT reset it; a server relying solely on
+        // comment heartbeats should not enable the idle timeout. A JSON decode
+        // error is per-event and non-fatal; every other error is terminal and
+        // stops the stream promptly.
+        let idle = self.sse_idle_timeout;
+        let guarded = unfold(
```
yoks, line 1285: why unfold here and not scan?

bmc-http/src/reqwest.rs
yoks: I think we still can address it by making SSL params their separate struct and reuse it in Client and ClientParams.

bmc-http/src/reqwest.rs
rohanharkin: Scans closure only runs when the stream yields an item where as unfold pulls the item and can wrap it in the timeout to detect if nothing arrives in time

bmc-http/src/reqwest.rs
rohanharkin: Ah okay yes, will do that now

bmc-http/src/reqwest.rs
yoks: I meant wrapping it with StreamExt, but it is are dependency, so lets keep it like this, it is clearer now.

bmc-http/src/reqwest.rs
rohanharkin: Clippy will still fail the CI on the client: ReqwestClient - I can keep the allow in or rename it what do you think is better

yoks: You can (and arguably should), reuse it for Client struct as well, as you copy same param there

bmc-http/src/reqwest.rs
```diff
@@ -488,6 +675,10 @@ impl Client {
         Self {
             client,
             retry: None,
+            sse: SseOptions {
```
yoks: You have SseOptions::default(), why not reuse it here?

bmc-http/src/reqwest.rs
```diff
@@ -389,6 +555,24 @@ impl ClientParams {
         self.retry = Some(retry);
         self
     }
+
+    /// Sets the maximum buffered size of a single, not-yet-terminated SSE event.
+    ///
+    /// See [`ClientParams::sse_max_event_bytes`].
```
yoks: This is no longer the case, it is SseOptions now

bmc-http/src/reqwest.rs
```diff
@@ -397,9 +581,11 @@ impl ClientParams {
 /// reqwest HTTP client library. It supports all standard HTTP features including
 /// TLS, authentication, and connection pooling.
 #[derive(Clone)]
+#[allow(clippy::struct_field_names)]
```
yoks: Ok i see, lint acts on client. Lets rename it client to inner. This is because it is wrapper on the ReqwestClient anyway.

yoks:
Looks good, but can you address few last nits? After that we can merge it.

yoks (approved):

PR #168, opened 2026-07-25
Fix: VeraRubin host BMC displays boot order in a different format

hakhondzadeh:
Redfish path redfish/v1/Systems/System_0 for a Vera Rubin host bmc, show boot orders in this format:
"BootOrder": [ "Boot0019: Ubuntu", "Boot0008: UEFI HTTPv4 (MAC:AC3AE2634C82)",
While previously, we saw them in this format:

 "BootOrder": [
            "Boot0002",
            "Boot0001",
            "Boot0000"
        ],
This PR fixes this issue so that boot order is properly added to exploration report.

### DIFF
--- a/redfish/src/computer_system/item.rs
+++ b/redfish/src/computer_system/item.rs
@@ -463,20 +463,6 @@ impl<B: Bmc> ComputerSystem<B> {
         }
     }
 
-    /// Returns whether `option` is referenced by a `BootOrder` entry on this system.
-    ///
-    /// Vera Rubin composite boot-order suffix handling is applied only when this BMC
-    /// is classified as Vera Rubin (`Vendor=NVIDIA`, `Product=VR NVL72`).
-    #[cfg(feature = "boot-options")]
-    #[must_use]
-    pub fn boot_option_matches_boot_order_entry(
-        &self,
-        option: &super::boot_option::BootOption<B>,
-        entry: BootOptionReference<&str>,
-    ) -> bool {
-        option.matches_boot_order_entry(entry, &self.bmc.quirks)
-    }
-
     /// Lenovo OEM extension
     ///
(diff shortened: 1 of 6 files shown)


redfish/src/computer_system/item.rs
```diff
@@ -463,6 +463,20 @@ impl<B: Bmc> ComputerSystem<B> {
         }
     }
 
+    /// Returns whether `option` is referenced by a `BootOrder` entry on this system.
+    ///
+    /// Vera Rubin composite boot-order suffix handling is applied only when this BMC
+    /// is classified as Vera Rubin (`Vendor=NVIDIA`, `Product=VR NVL72`).
+    #[cfg(feature = "boot-options")]
+    #[must_use]
+    pub fn boot_option_matches_boot_order_entry(
```
poroh: This solution is not ideal because it requires user to know that this is the only way to match boot option to boot order entry. I would suggest to add patch support for ComputerSystem item and remove ": <DisplayName>" before deserialization to  ComputerSystemSchema.

All user logic will remain the same as today.

To do this you need to introduce additional patch function similar to:
https://github.com/NVIDIA/nv-redfish/blob/c62a65a2477356c5f54c67ca9acf0e9a30562c06/redfish/src/computer_system/mod.rs#L114

You can find couple similar functions nearby.

poroh (changes requested):
See my comment.

yoks:
Also, if we find bug in the BMC implementation (like this one). We need to file ticket for the vendor. Otherwise this quirks will live forever

redfish/src/computer_system/item.rs
hakhondzadeh: Thanks, changed the implementation. 
please take another look.

poroh (approved):

PR #180, opened 2026-08-01
feat: expose network adapter port MAC addresses

### LINKED ISSUE #176: Expose Network Adapter Port MAC Addresses
Give `nv-redfish` callers typed access to network-adapter ports and their MAC addresses, including Lenovo XCC's OEM-only physical-port value. This lets inventory clients discover installed host NICs even when `Systems/.../EthernetInterfaces` is empty.

## What this involves

- Add a `ports` feature for the standard `PortCollection` and `Port` schemas.
- Expose `NetworkAdapter::ports()` and typed collection/resource wrappers.
- Read `Port.Ethernet.AssociatedMACAddresses`, falling back to `Oem.Lenovo.PhysicalPortMacAddress` when the standard values are absent or empty.
- Keep standard values authoritative and Lenovo-specific handling internal to `nv-redfish`.
- Return the existing typed MAC-address wrappers without changing the server-provided format.
- Add tests for standard, Lenovo-only, precedence, empty-value, and collection-traversal behavior.

This supports https://github.com/NVIDIA/infra-controller/issues/4436

chet:
We ran into a Lenovo host with an empty `Systems/.../EthernetInterfaces` collection, which meant `site-explorer` couldn't find the host MAC it needed to recognize the PXE interface. The NIC was there -- it was just exposed under `Chassis/.../NetworkAdapters/.../Ports`, with Lenovo XCC putting the physical-port address in `Oem.Lenovo.PhysicalPortMacAddress`.

I don't think every caller should have to learn that Lenovo detail. `nv-redfish` is our vendor abstraction layer, so this teaches it to traverse the standard port collection and fall back to Lenovo's OEM value only when `Port.Ethernet.AssociatedMACAddresses` is missing or empty. That gives callers one path for both cases, while standard values still win and the MAC text stays exactly as the BMC returned it.

## What changed

- add typed `PortCollection` and `Port` wrappers behind a new `ports` feature
- expose `NetworkAdapter::ports()` for standard adapter-port traversal
- prefer standard associated MAC addresses and fall back to Lenovo XCC's OEM physical-port value
- cover standard, Lenovo-only, precedence, empty-value, missing-link, and end-to-end collection behavior

## Testing

- `make ci`
- `cargo test -p nv-redfish-tests --test test-network-adapter-port -- --no-capture`

## Related issues

This supports https://github.com/NVIDIA/nv-redfish/issues/176.

This supports https://github.com/NVIDIA/infra-controller/issues/4436.

Closes #176


redfish/src/port.rs
```diff
@@ -0,0 +1,140 @@
...
+}
+
+impl<B: Bmc> Port<B> {
+    pub(crate) async fn new(
+        bmc: &NvBmc<B>,
+        nav: &NavProperty<PortSchema>,
+    ) -> Result<Self, Error<B>> {
+        let data = nav.get(bmc.as_ref()).await.map_err(Error::Bmc)?;
+        #[cfg(feature = "oem-lenovo")]
+        let oem_lenovo = LenovoPort::new(&data).map_err(Error::Json)?;
```
jayzhudev: Should Lenovo port parsing be deferred until a fallback is needed? With this feature enabled, an incompatible `Lenovo` value can fail `Port::new()` and the whole collection. But valid standard MAC address data may be available in the meantime.

redfish/src/oem/lenovo/mod.rs
```diff
@@ -29,5 +29,9 @@ pub mod security_service;
 #[cfg(feature = "computer-systems")]
 pub mod computer_system;
 
+/// Support of Lenovo Port OEM attributes.
+#[cfg(feature = "ports")]
+pub(crate) mod port;
```
yoks: why pub(crate) here? Is not used anywhere inside.

redfish/src/port.rs
```diff
@@ -0,0 +1,140 @@
...
+/// Network port collection.
+///
+/// Provides functions to access collection members.
+pub struct PortCollection<B: Bmc> {
+    bmc: NvBmc<B>,
+    collection: Arc<PortCollectionSchema>,
+}
+
+impl<B: Bmc> PortCollection<B> {
+    #[allow(dead_code)] // Used by NetworkAdapter when that feature is enabled.
```
yoks: I think you can drop it, it would be generated only when port is reachable which requires network adapters

redfish/src/port.rs
```diff
@@ -0,0 +1,140 @@
...
+
+/// Network port collection.
+///
+/// Provides functions to access collection members.
+pub struct PortCollection<B: Bmc> {
+    bmc: NvBmc<B>,
+    collection: Arc<PortCollectionSchema>,
+}
+
+impl<B: Bmc> PortCollection<B> {
```
yoks: we usually spliting collection and port in different files in repo. Rule of thumb one primary struct per rs file

redfish/src/port.rs
yoks: Yes even worse whole `members()` would fail if even one fails here.

redfish/src/oem/lenovo/port.rs
```diff
@@ -0,0 +1,57 @@
...
+                serde_json::from_value(data.clone()).map(|data| Self {
+                    data: Arc::new(data),
+                })
+            })
+            .transpose()
+    }
+
+    /// Get Lenovo's physical-port MAC address.
+    #[must_use]
+    pub fn physical_port_mac_address(&self) -> Option<&str> {
```
yoks: why erase type here?

redfish/src/port.rs
```diff
@@ -0,0 +1,140 @@
...
+
+    /// Get MAC addresses associated with this port.
+    ///
+    /// Standard `Ethernet.AssociatedMACAddresses` values are authoritative.
+    /// Lenovo XCC reports the physical address only through
+    /// `Oem.Lenovo.PhysicalPortMacAddress` on some systems, so that value is
+    /// returned when the standard array is absent or empty.
+    ///
+    #[must_use]
+    pub fn associated_mac_addresses(&self) -> Vec<MacAddress<'_>> {
```
yoks: This function doing a little too much. It is not clear from outside (by name at least), what it OEM dependent. `associated_mac_addresses` is something standart, this silently switches to Lenovo if it not found and Lenovo OEM enabled.

If we want something with vendor fallback (something smart) we need name it as such i think. 

It also kinda pushes Site Explorer logic into nv-redfish, which i'm not fan of, better be other way around. But i will let @poroh to chime here.

poroh: nv-redfish rule of thumb is "do not try to be smart when provide data to library user". If data is available through OEM extension then nv-redfish should provide access to OEM extension and let library user to choose if they want to use LenovoPort::physical_port_mac_address as  associated_mac_addresses.

This decision is not nv-redfish buisness.

poroh (changes requested):

redfish/src/port.rs
chet: Yeah, good call -- I’ll move Lenovo parsing out of `Port::new()` and defer it to explicit OEM access, so a malformed Lenovo payload can’t fail `members()` or hide usable standard MAC data. I'll post back when it's done, lemme tweak.

redfish/src/oem/lenovo/mod.rs
chet: Sooo, this is/was `pub(crate)` because `port` wasn't visible to `oem::lenovo`. BUT, @poroh's comment below will actually change this, with a pub `LenovoPort` that we'll make available to  through `Port::oem_lenovo()`. I'll ping back once it's updated! Looking through the other comments still.

redfish/src/port.rs
chet: Nice okay! Will drop.

redfish/src/port.rs
chet: Okay nice -- yeah I like that. Give me a bit; I'll split this across port/mod.rs, port/collection.rs, and port/item.rs.

redfish/src/oem/lenovo/port.rs
chet: Yupppp -- will update to return `Option<MacAddress>`.

redfish/src/port.rs
chet: Yeah agreed. I realize we're fixing a lot of anti-patterns we inherited and/or continued with `libredfish` with this repo -- thanks for ensuring it stays that way. I'll fiddle with it so `port.associated_mac_addresses()` returns just standard `Ethernet.AssociatedMACAddresses`, and then `port.oem_lenovo()` returns `LenovoPort`, so site-explorer would end up doing something like...

```
  let addresses = port.associated_mac_addresses();
  if addresses.is_empty() {
      if let Some(lenovo) = port.oem_lenovo()? {
          if let Some(address) = lenovo.physical_port_mac_address() {
              // and then site-explorer does some fallback work here or something like that
          }
      }
  }
  ```
  
  Back in a bit!

File: redfish/src/port/item.rs [hunk 7]
```diff
@@ -0,0 +1,93 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::mac_address::MacAddress;
+#[cfg(feature = "oem-lenovo")]
+use crate::oem::lenovo::port::LenovoPort;
```
poroh: nitpick: I prefer to separate conditional imports and non-conditional imports.

poroh (approved):

jayzhudev (approved):
LGTM

redfish/src/port/item.rs
chet: @poroh Updated! I think I'll need a re-approval now though.

poroh (approved):

yoks (approved):

jayzhudev (approved):
Sending extra love : )

Issue #184, opened 2026-08-04
Preserve BlueField OEM and Port Collection Data


chet:
nv-redfish callers should keep the BlueField OEM data and valid NetworkAdapter Ports they need when inventory is only partially usable. That lets infra-controller adopt v0.13 without losing BF3 Mode/BaseMAC or discarding every good Port after one member fails.

## What this involves

- Classify the supported `Vendor=Nvidia`, `Product=BlueField-3 DPU` service-root identity as the existing NVIDIA DPU quirk platform so `Oem.Nvidia` keeps its forced fetch.
- Add a non-breaking `PortCollection::member_results()` path that reports each referenced member independently and identifies failed members.
- Preserve `PortCollection::members()` short-circuiting behavior for existing callers.
- Cover BF3 OEM dereferencing and partial Port success.

This supports https://github.com/NVIDIA/infra-controller/issues/4469

PR #185, opened 2026-08-04
fix: preserve BlueField OEM and port collection data

chet:
nv-redfish callers should be able to keep the BF3 OEM inventory and valid NetworkAdapter Ports they need when the surrounding Redfish inventory is only partially usable.

The v0.13 OEM fetch path depends on NVIDIA DPU platform detection, but infra-controller's BF3 service-root identity uses `Vendor=Nvidia`, `Product=BlueField-3 DPU`. Without recognizing that signature, the linked OEM payload is not fetched and its `BaseMAC` and `Mode` disappear. This extends the existing NVIDIA DPU quirk to that exact BF3 product identity.

Port collections have a similar all-or-nothing problem for callers that want best-effort inventory: `members()` stops at the first failed member and cannot return earlier successes. This adds a separate `member_links()` API using `EntityLink`, allowing consumers to fetch and handle each Port independently while leaving the existing `members()` behavior unchanged.

Tests added!

This supports https://github.com/NVIDIA/nv-redfish/issues/184

This also supports https://github.com/NVIDIA/infra-controller/issues/4469

### DIFF
--- a/redfish/src/port/collection.rs
+++ b/redfish/src/port/collection.rs
@@ -14,6 +14,7 @@
 // limitations under the License.
 
 use super::Port;
+use super::PortLink;
 use crate::schema::port_collection::PortCollection as PortCollectionSchema;
 use crate::Error;
 use crate::NvBmc;
@@ -52,5 +53,18 @@ impl<B: Bmc> PortCollection<B> {
             members.push(Port::new(&self.bmc, member).await?);
         }
         Ok(members)
+    }
+
+    /// Return lazy links for the ports in this collection.
+    ///
+    /// Each link can be fetched independently, allowing callers to choose their own error-handling
+    /// policy without changing the eager, all-or-nothing behavior of [`Self::members`].
+    #[must_use]
+    pub fn member_links(&self) -> Vec<PortLink<B>> {
+        self.collection
+            .members
+            .iter()
+            .map(|member| PortLink::new(&self.bmc, NavProperty::new_reference(member.id().clone())))
+            .collect()
     }
 }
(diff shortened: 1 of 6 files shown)


redfish/src/port/collection.rs
```diff
@@ -53,4 +54,22 @@ impl<B: Bmc> PortCollection<B> {
         }
         Ok(members)
     }
+
+    /// Fetch every port in this collection without discarding successful members.
+    ///
+    /// Results appear in server order, with failed members carrying their identifier. This lets
+    /// callers retain valid ports when another referenced member cannot be fetched or parsed.
+    /// Errors decoding an already-expanded collection are returned while fetching the collection.
+    pub async fn member_results(&self) -> Vec<Result<Port<B>, (ODataId, Error<B>)>> {
```
yoks: This is good API question. We eagerly fetch members, but maybe we can return links instead and leave the filtering of errors to the end users? This is more usable, so i would do member_links, which would return port ODATA which consumer fetch themselves, rather than have it here.

yoks: E.g. you can look at collect_sensors which return NavProperty. Or EntityLink

chet: Yeah, I like that. Returning links keeps the fetch/error handling with the consumer, while preserving the existing members() API for callers that want that specific existing behavior.

 I’ll switch `member_results()` to `member_links()` using `EntityLink`, and add the corresponding `PortLink`/upgrade support -- gimme a few. If you want me to tweak anything else, happy to tweak!

yoks (approved):

poroh (approved):

PR #186, opened 2026-08-05
feat: csdl query API

yoks:
Public API for CSDL Schema with ability to query it.

Can be used by other libs (e.g. https://github.com/NVIDIA/nv-telemetry) for code generation based on nv-redfish schemas

### DIFF
--- a/csdl-compiler/src/query.rs
+++ b/csdl-compiler/src/query.rs
@@ -16,7 +16,6 @@
 //! Schema queries for projection compilers.
 
 use std::collections::HashMap;
-use std::iter;
 
 use crate::compiler::Compiled;
 use crate::compiler::Config;
@@ -65,9 +64,9 @@ impl<'a> SchemaQuery<'a> {
         let mut entities = HashMap::new();
         for qname in compiled.entity_types.keys() {
             let name = qname.name.inner().as_str();
-            let replace = entities.get(name).is_none_or(|current| {
-                chain_length(&compiled, *qname) > chain_length(&compiled, *current)
-            });
+            let replace = entities
+                .get(name)
+                .is_none_or(|current| Rank::new(*qname) > Rank::new(*current));
             if replace {
                 entities.insert(name, *qname);
             }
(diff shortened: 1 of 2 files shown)


csdl-compiler/src/query.rs
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    /// Resolves a dotted property path against the named entity type,
+    /// descending through complex types and walking base chains.
+    #[must_use]
+    pub fn resolve(&self, entity: &str, path: &str) -> Option<ResolvedProperty<'a>> {
+        let mut current = TypeRef::Entity(*self.entities.get(entity)?);
+        let mut segments = path.split('.').peekable();
+        loop {
+            let segment = segments.next()?;
```
jayzhudev: Maybe replace`loop` with:

```rust
while let Some(segment) = segments.next() ...
```

When I see `loop` my alertness goes up even if it's correct 😁

csdl-compiler/src/query.rs
```diff
@@ -0,0 +1,264 @@
...
+    }
+}
+
+#[derive(Clone, Copy)]
+enum TypeRef<'a> {
+    Entity(QualifiedName<'a>),
+    Complex(QualifiedName<'a>),
+}
+
+fn chain_length<'a>(compiled: &Compiled<'a>, mut qname: QualifiedName<'a>) -> usize {
```
jayzhudev: LLM suggested this can be refactored to:

```rust
fn chain_length<'a>(compiled: &Compiled<'a>, start: QualifiedName<'a>) -> usize {
    std::iter::successors(Some(start), |&q| compiled.entity_types.get(&q)?.base).count()
 }
```

I don't have a preference. The while loop makes it easy to read, plus this is not runtime code. The one-liner doesn't look that bad too.

csdl-compiler/src/query.rs
```diff
@@ -0,0 +1,264 @@
...
+                };
+                return Some(ResolvedProperty {
+                    type_name: qname,
+                    class,
+                    nullable: property.nullable.into_inner(),
+                    collection,
+                    enum_members,
+                });
+            }
+            if class != TypeClass::ComplexType {
```
jayzhudev, line 111: Would using `matches!` be more idiomatic?

csdl-compiler/src/query.rs
```diff
@@ -0,0 +1,264 @@
...
+    ///
+    /// Returns a compile error if any type cannot be resolved.
+    pub fn build(bundle: &'a SchemaBundle) -> Result<Self, Error<'a>> {
+        let config = Config {
+            entity_type_filter: EntityTypeFilter::new_restrictive(Vec::new()),
+            ..Config::default()
+        };
+        let compiled = bundle.compile_all(config)?;
+
+        let mut entities: HashMap<&str, QualifiedName<'_>> = HashMap::new();
```
jayzhudev: Maybe the HashMap types can be inferred by the insert down below? I didn't try but it seems possible.

jayzhudev (approved):
LGTM, some nits.

csdl-compiler/src/query.rs
yoks: Matches is fine for complex types, but this is simple enum, i think simple != is better here.

csdl-compiler/src/query.rs
yoks: It can, i do not have preference, but explicit type is a little more redable for something complex as this one.

csdl-compiler/src/query.rs
```diff
@@ -0,0 +1,259 @@
...
+            }
+            current = TypeRef::Complex(qname);
+        }
+        None
+    }
+
+    /// The named structural property, searched up the base chain.
+    fn property_of(&self, mut tref: TypeRef<'a>, name: &str) -> Option<&Property<'a>> {
+        loop {
+            let (properties, base) = self.declaration(tref)?;
```
jayzhudev: Same here:

```rust
while let Some((properties, base)) = self.declaration(tref) {...
```

jayzhudev (approved):
I forgot to click the button for this one.

jayzhudev (approved):

jayzhudev (approved):

poroh (approved):

poroh (approved):

jayzhudev (approved):

PR #191, opened 2026-08-08
feat: support deprecated fields

yoks:
CSDL now parses deprecated fields and mark them as optional

**This is the breaking change**

### DIFF
--- a/core/src/nav_property.rs
+++ b/core/src/nav_property.rs
@@ -205,6 +205,17 @@ impl<T: EntityTypeRef> NavProperty<T> {
             Self::Expanded(v) => v.0.odata_id(),
         }
     }
+
+    /// The property value, if the property was expanded in the
+    /// response. Unlike [`NavProperty::get`] this never touches the
+    /// BMC: a reference variant yields `None`.
+    #[must_use]
+    pub fn expanded(&self) -> Option<&T> {
+        match self {
+            Self::Expanded(v) => Some(v.0.as_ref()),
+            Self::Reference(_) => None,
+        }
+    }
 }
 
 impl<T: EntityTypeRef + for<'de> Deserialize<'de> + 'static> NavProperty<T> {
--- a/csdl-compiler/src/edmx/annotation.rs
+++ b/csdl-compiler/src/edmx/annotation.rs
@@ -18,10 +18,7 @@
 use crate::edmx::attribute_values::Error as AttributeValuesError;
 use crate::edmx::EnumMemberName;
 use crate::edmx::QualifiedTypeName;
-use serde::de::Error as DeError;
-use serde::de::Visitor;
 use serde::Deserialize;
-use serde::Deserializer;
 use std::fmt::Display;
 use std::fmt::Formatter;
 use std::fmt::Result as FmtResult;
@@ -39,12 +36,28 @@ pub struct Annotation {
     pub bool_value: Option<bool>,
     #[serde(rename = "@Int")]
     pub int_value: Option<i64>,
+    /// 14.4.7 `EnumMember` expression, kept verbatim. Use
+    /// [`Annotation::enum_members`].
     #[serde(rename = "@EnumMember")]
-    pub enum_member: Option<Box<AnnotationEnumMember>>,
+    enum_member: Option<String>,
     #[serde(rename = "Collection")]
     pub collection: Option<AnnotationCollection>,
     #[serde(rename = "Record")]
     pub record: Option<AnnotationRecord>,
+}
+
+impl Annotation {
+    /// The `EnumMember` expression members (14.4.7).
+    ///
+    /// A value may hold several space-separated members (`IsFlags`
+    /// enumerations); they are yielded individually, in document
+    /// order. Members outside the `Namespace.Type/Member` grammar of
+    /// [`AnnotationEnumMember`] are skipped: the document is stored
+    /// verbatim and interpretation is left to the consumer, so an
+    /// unsupported member cannot fail the parse of the document.
+    pub fn enum_members(&self) -> impl Iterator<Item = AnnotationEnumMember> + '_ {
+        parse_enum_members(self.enum_member.iter())
+    }
 }
 
 #[derive(Debug, Deserialize)]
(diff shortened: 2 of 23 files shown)


csdl-compiler/src/edmx/annotation.rs
```diff
@@ -82,12 +95,52 @@ pub struct PropertyValue {
...
+// Parse the members of EnumMember expression values, skipping members
+// that do not follow the AnnotationEnumMember grammar: consumers
+// search these members for specific values, and an unsupported member
+// cannot be the value searched for.
+fn parse_enum_members<'a>(
+    values: impl Iterator<Item = &'a String> + 'a,
+) -> impl Iterator<Item = AnnotationEnumMember> + 'a {
+    values
+        .flat_map(|value| value.split_whitespace())
+        .filter_map(|member| member.parse().ok())
```
jayzhudev, line 136: LLM: enum member parsing now fails open

- Valid unknown members, such as `Future.RevisionKind/NewValue`, may be ignored.
- Invalid syntax, such as `Redfish.RevisionKind/Deprecated/Garbage`, should fail parsing.

jayzhudev (approved):
LGTM. Thanks for this change!

nit: PR title has a typo
nit: need to note breaking change for the next release

redfish/src/bmc_quirks.rs
```diff
@@ -62,16 +66,22 @@ impl BmcQuirks {
...
             None if redfish_version_str == Some("1.9.0") => Some(Platform::Anonymous1_9_0),
             _ => None,
         };
-        Self { platform }
+        Self {
+            platform,
+            // Both vendor string casings are in the wild: "NVIDIA" on
+            // host BMCs and "Nvidia" on BlueField DPUs.
+            #[cfg(feature = "event-service")]
+            nvidia_vendor: matches!(vendor_str, Some("NVIDIA" | "Nvidia")),
```
lauragra-y, line 83: This excludes Wiwynn ODM GB200 NVL trays, which report `"Vendor": "WIWYNN"`. Adding "WIWYNN" to the matches! would cover it.
See bug 6579817.

lauragra-y (changes requested):

redfish/src/bmc_quirks.rs
lauragra-y: Agreed to add this as a separate PR after this one merges.

jayzhudev (approved):

lauragra-y (approved):

core/src/nav_property.rs
```diff
@@ -205,6 +205,17 @@ impl<T: EntityTypeRef> NavProperty<T> {
             Self::Expanded(v) => v.0.odata_id(),
         }
     }
+
+    /// The property value, if the property was expanded in the
+    /// response. Unlike [`NavProperty::get`] this never touches the
+    /// BMC: a reference variant yields `None`.
+    #[must_use]
+    pub fn expanded(&self) -> Option<&T> {
```
poroh, line 213: This method is only added for tests. Should we introduce it it here as part of support of deprecation?

csdl-compiler/src/edmx/annotation.rs
```diff
@@ -39,14 +36,30 @@ pub struct Annotation {
     pub bool_value: Option<bool>,
     #[serde(rename = "@Int")]
     pub int_value: Option<i64>,
+    /// 14.4.7 `EnumMember` expression, kept verbatim. Use
+    /// [`Annotation::enum_members`].
     #[serde(rename = "@EnumMember")]
-    pub enum_member: Option<Box<AnnotationEnumMember>>,
+    enum_member: Option<String>,
```
poroh, line 42: This change and change below looks strange to me. If more than one member can present in this field then the type here should represent this fact. Also, I don't think that we should delay parsing until enum_members() call from compiler.

csdl-compiler/src/generator/rust/struct_def.rs
```diff
@@ -75,6 +76,7 @@ pub struct StructDef<'a> {
     // it in active resource itself.
     need_redfish_settings: bool,
     dynamic_properties: Option<DynamicProperties<'a>>,
+    deprecation: Option<Box<Deprecation>>,
```
poroh, line 79: Do we have reason to clone it to this struct? Maybe &'a Deprecation is enough?

csdl-compiler/src/redfish/mod.rs
```diff
@@ -36,3 +36,13 @@ pub struct DynamicProperties<'a> {
     pub pattern: &'a String,
     pub ptype: &'a String,
 }
+
+/// Deprecation revision of a property: a `Redfish.Revisions` record
+/// with `Kind == Redfish.RevisionKind/Deprecated`.
+#[derive(Debug, Clone)]
+pub struct Deprecation {
```
poroh, line 43: Most of properties here contains references to the initial parsed EDMX document. Please consider to use  
```
struct Deprecation<'a> {
   version: Option<&'a str>
   description: Option<'a str>
}
```

redfish/src/bmc_quirks.rs
```diff
@@ -23,6 +23,12 @@ use crate::account::SlotDefinedConfig as SlotDefinedUserAccountsConfig;
 /// for each individual platform class.
 pub struct BmcQuirks {
     platform: Option<Platform>,
+    /// Vendor is NVIDIA regardless of which platform class the
+    /// product was classified into. Quirks that apply to the whole
+    /// vendor family gate on this instead of enumerating platforms,
+    /// so a new product-specific match arm cannot silently drop them.
+    #[cfg(feature = "event-service")]
+    nvidia_vendor: bool,
```
poroh, line 31: I don't think that we should add "nvidia_vendor". If we need to apply quirks for all NVIDIA products for some reasons we should have function is_nvidia_vendor() for the platform field above.

redfish/src/bmc_quirks.rs
```diff
@@ -250,3 +260,42 @@ impl BmcQuirks {
         )
     }
 }
+
+#[cfg(test)]
+#[cfg(feature = "event-service")]
+mod tests {
```
poroh, line 266: These test is out of place. All checks for individual vendors support are in integration tests for corresponding feature

poroh (changes requested):

PR #202, opened 2026-08-21
feat(quirks): add Dell PowerEdge XE9780 platform class

shevchenko-evgeny:
The iDRAC service root reports only Vendor: "Dell", which the XE9780 shares with every other PowerEdge, so it cannot be told apart from the root payload alone. Classification now walks the computer system collection and looks for a member whose Model contains xe9780. The new platform inherits the existing Dell quirks and adds two of its own:
- $expand responses are incomplete, so collections are fetched with plain GETs and members individually.
- chassis UUID may carry an arbitrary string rather than a GUID; a read patch replaces such values with null so the resource still deserializes.
relates to: [NICO ticket](https://github.com/NVIDIA/infra-controller/issues/3227)

### DIFF
--- a/redfish/src/bmc_quirks.rs
+++ b/redfish/src/bmc_quirks.rs
@@ -13,7 +13,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use nv_redfish_core::Bmc;
+
 use crate::schema::service_root::ServiceRoot;
+use crate::Error;
 
 #[cfg(feature = "accounts")]
 use crate::account::SlotDefinedConfig as SlotDefinedUserAccountsConfig;
@@ -27,10 +30,11 @@ pub struct BmcQuirks {
 
 // Platform shouldn't be considered as vendor. Actually it is class of
 // devices that have the same set of quirks.
-#[derive(PartialEq, Eq)]
+#[derive(PartialEq, Eq, Debug)]
 enum Platform {
     Hpe,
     Dell,
+    DellPEXE9780,
     AmiViking,
     AmiGb300,
     VeraRubin,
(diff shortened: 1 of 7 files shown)


redfish/src/bmc_quirks.rs
```diff
@@ -42,8 +46,9 @@ enum Platform {
 }
 
 impl BmcQuirks {
-    pub fn new(root: &ServiceRoot) -> Self {
+    pub async fn new<B: Bmc>(root: &ServiceRoot, bmc: &B) -> Result<Self, Error<B>> {
```
yoks, line 49: This is quite a change. And it shows gaps in quirk platform implementation (maybe).

So i would rather think on how properly support such cases. I would not merge it as is, as it looks more like a very ugly crutch and design problem

shevchenko-evgeny: @yoks @poroh Could you please advise what potential changes could get this accepted?
I see your concerns, but I have no idea except to apply the quirks to the whole Dell fleet. My point is that it is better to be as targeted as possible. Honestly, I would even think about adding a BMC software version check before applying any quirks. A combination of Vendor, Model, and FW version looks reasonable to me. What do you think?

williampnvidia: I agree model targeting is preferable, but the current detection puts network I/O and feature dependencies inside BmcQuirks::new. It also deserializes ComputerSystem before the existing Dell patches are available so a known malformed LastResetTime can make member.get() fail and silently classify an XE9780 as generic Dell, leaving the new quirks disabled. Adding firmware matching would not solve that bootstrap problem and may be brittle unless the behavior is proven to be version specific. 

Short term, could the safe invalid-UUID normalization, and possibly plain GET behavior, apply to Dell generally? Longer term, platform identity could be collected separately and passed to a non-I/O quirk resolver.

redfish/src/bmc_quirks.rs
```diff
@@ -75,7 +83,38 @@ impl BmcQuirks {
...
+        }
+        Ok(false)
+    }
+
+    #[cfg(not(feature = "computer-systems"))]
+    async fn check_if_dell_pe_x9780<B: Bmc>(
+        _root: &ServiceRoot,
+        _bmc: &B,
+    ) -> Result<bool, Error<B>> {
+        Err(Error::MissingFeature("computer-systems".into()))
```
williampnvidia, line 117: With computer-systems disabled, this helper is still called for every Dell BMC and returns an error, so ServiceRoot::new now fails for valid feature sets such as chassis, accounts, or update-service. Please fall back to generic Dell behavior when model probing is unavailable or return Ok(false) rather than treating an optional feature as fatal.

PR #203, opened 2026-08-21
fix(bmc-http): derive cache validators from the cached body

shevchenko-evgeny:
`get_with_cache` kept ETags in a `HashMap` separate from the response-body cache, but served 304 responses from that body cache. The two could disagree: while one request waits for its HTTP response, another can evict its body, so the 304 arrives with nothing to serve and the GET fails with `CacheMiss` on a resource that was perfectly fetchable.
relates to [PE XE9780 support](https://github.com/NVIDIA/infra-controller/issues/3227) and [SMC GB300 NVL support](https://github.com/NVIDIA/infra-controller/issues/4096)

### DIFF
--- a/bmc-http/src/lib.rs
+++ b/bmc-http/src/lib.rs
@@ -49,7 +49,6 @@ mod schema;
 #[cfg(feature = "reqwest")]
 pub mod reqwest;
 
-use std::collections::HashMap;
 use std::error::Error as StdError;
 use std::fmt;
 use std::future::Future;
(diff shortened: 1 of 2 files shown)


bmc-http/src/lib.rs
```diff
@@ -582,17 +579,23 @@ where
         // The `etag` is always `None` when caching is disabled. Check the flag here so we can save
         // a read lock acquisition and guarantee that disabled caching never sends If-None-Match,
```
yoks, line 580: Need update this comment as it write lock, also while it seems cheap it worth mentioned we gettitng write lock on cache now.

yoks (approved):

bmc-http/src/lib.rs
yoks: Would not hold this PR on this tho. Will land in later PR.

PR #204, opened 2026-08-30
feat(redfish)!: serialize generated resource models

jayzhudev:
Add opt-in `serde` serialization for generated Redfish read and excerpt models and their supporting action and navigation types. Applications can serialize resources such as `LogEntry` for content fingerprinting, persistence, and other downstream use.

Enable generated model serialization with the `resource-serialization` feature.

### Breaking Changes

This is a source-breaking change for callers that construct `generator::rust::Config` with a struct literal.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,7 @@ std-not-standalone-features = assembly \
 
 std-standalone-features = $(filter-out $(std-not-standalone-features),$(all-std-features))
 
-ci-features-list := $(subst $(space),$(comma),$(all-std-features)),http-extras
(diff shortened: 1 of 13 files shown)


poroh (changes requested):
Serialization derivation generates 2x code for each individual type. For many purposes this code is not required at all. So, in my opinion this feature should be opt-in.

Also, I would suggest to consider alternative opt-in feature of nv-redfish instead of this one -- provide access to original JSON payload.

poroh (approved):

PR #208, opened 2026-09-03
feat(redfish): prefer NVIDIA DPU reset for force restart

zanezhuNVIDIA:
Using the "/redfish/v1/Chassis/BlueField_0/Actions/Oem/NvidiaChassis.Reset" endpoint is the official way for BF4 to do the reset.
The standard redfish ForceRestart method may not work with BF4 DPU.
To handle this situation, the nv-redfish would try to use "/redfish/v1/Chassis/BlueField_0/Actions/Oem/NvidiaChassis.Reset" endpoint first to reset the BF4 DPU if the endpoint exists. If the endpoint does not exist, the nv-redfish would be fallback to standard ForceRestart method.

## Summary by CodeRabbit

* **New Features**
  * Added NVIDIA BlueField-4 DPU reset support through the chassis OEM reset action.
  * Computer system restarts fall back to the standard reset action when the NVIDIA-specific action is unavailable or not found.
* **Bug Fixes**
  * Preserved chassis information during computer-system updates to keep NVIDIA reset handling available.
  * Improved recognition of unavailable reset actions and missing action targets.
* **Tests**
  * Added coverage for NVIDIA-specific resets and standard reset fallback scenarios.

### DIFF
--- a/redfish/src/chassis/item.rs
+++ b/redfish/src/chassis/item.rs
@@ -222,7 +222,7 @@ impl<B: Bmc> Chassis<B> {
             return Ok(None);
         };
 
-        action
+        match action
             .run(
                 self.bmc.as_ref(),
                 &ForceDpuResetRequest {
@@ -230,8 +230,11 @@ impl<B: Bmc> Chassis<B> {
                 },
             )
             .await
-            .map(Some)
-            .map_err(Error::Bmc)
+        {
+            Ok(response) => Ok(Some(response)),
+            Err(error) if nv_redfish_core::ActionError::is_not_found(&error) => Ok(None),
+            Err(error) => Err(Error::Bmc(error)),
+        }
     }
 
     /// Get hardware identifier of the network adpater.
--- a/redfish/src/chassis/item.rs
+++ b/redfish/src/chassis/item.rs
@@ -68,6 +68,23 @@ use crate::sensor::SensorLink;
 #[cfg(feature = "oem-nvidia")]
 use std::convert::identity;
 
+#[cfg(all(feature = "oem-nvidia", feature = "computer-systems"))]
+const NVIDIA_RESET_ACTION: &str = "#NvidiaChassis.Reset";
+
+#[cfg(all(feature = "oem-nvidia", feature = "computer-systems"))]
+#[derive(serde::Serialize)]
+struct ForceDpuResetRequest {
+    #[serde(rename = "ResetType")]
+    reset_type: NvidiaChassisResetType,
+}
+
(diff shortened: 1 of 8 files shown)


redfish/src/computer_system/item.rs
```diff
@@ -201,6 +218,13 @@ impl<B: Bmc> ComputerSystem<B> {
     where
         B::Error: nv_redfish_core::ActionError,
     {
+        #[cfg(all(feature = "chassis", feature = "oem-nvidia"))]
+        if reset_type == Some(ResetType::ForceRestart) {
+            if let Some(response) = self.try_oem_nvidia_force_dpu_reset().await? {
```
poroh, line 223: nv-redfish is designed to be buisness-logic agnostic. It only provides building blocks for anybody who wants to implement any buisness logic on top of it.

Rule "use chassis reboot instead of computer system" is buisness logic to support BF4. So, nv-redfish is not right place for it. nv-redifsh only should provide OEM reset from chassis object.

NICo should this in a way you defined below.

redfish/src/chassis/item.rs
```diff
@@ -176,6 +193,47 @@ impl<B: Bmc> Chassis<B> {
...
+        action
+            .run(
+                self.bmc.as_ref(),
+                &ForceDpuResetRequest {
+                    reset_type: NvidiaChassisResetType::ForceDpuReset,
+                },
+            )
+            .await
+            .map(Some)
+            .map_err(Error::Bmc)
```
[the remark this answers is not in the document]
zanezhuNVIDIA: Fixed in 4b43441. A Not Found response from the advertised NVIDIA reset action now falls back to the standard ComputerSystem.Reset action, while other BMC errors are preserved. Added regression coverage for the stale action target.

redfish/src/chassis/item.rs
```diff
@@ -68,6 +68,23 @@ use crate::sensor::SensorLink;
...
+#[cfg(all(feature = "oem-nvidia", feature = "computer-systems"))]
+#[derive(serde::Serialize)]
+struct ForceDpuResetRequest {
+    #[serde(rename = "ResetType")]
+    reset_type: NvidiaChassisResetType,
+}
+
+#[cfg(all(feature = "oem-nvidia", feature = "computer-systems"))]
+#[derive(serde::Serialize)]
+enum NvidiaChassisResetType {
```
poroh, line 83: I'll create PR that will enable generation of this type from CSDL schema.

poroh (changes requested):

PR #215, opened 2026-09-11
feat: SSE events carry last seen message

yoks:
Before API did not allowed to resume SSE streams fron the last seen id. This PR adds api `sse_events` which allows to supply this information. 

This is non breaking change (hence new function)

## Summary by CodeRabbit

- **New Features**
  - Added Server-Sent Event streams with event ID tracking and resumption from a specified event ID.
  - Added Event Service APIs for resuming streams from a selected event.
  - Preserved compatibility with streams that do not include event IDs.
  - Added mock stream support for scripted events and resumption scenarios.

- **Bug Fixes**
  - Improved SSE frame handling, timeout behavior, and error processing.
  - Ensured event stream requests honor concurrency limits during connection setup.

### DIFF
--- a/bmc-http/src/reqwest.rs
+++ b/bmc-http/src/reqwest.rs
@@ -1313,11 +1313,16 @@ impl HttpClient for Client {
         custom_headers: &HeaderMap,
         last_event_id: Option<&str>,
     ) -> Result<BoxTryStream<StreamEvent<T>, Self::Error>, Self::Error> {
+        // An empty id is no id, as the SSE processing model has it: none is
+        // sent, and none is in effect until the server sets one. Otherwise
+        // the resume id stays in effect across the reconnection, so events
+        // the server sends without an `id` field still carry it.
+        let last_event_id = last_event_id.filter(|id| !id.is_empty());
         let frames = self
             .sse_frames(url, credentials, custom_headers, last_event_id)
             .await?;
         let events = frames
-            .scan(None, |last_event_id, frame| {
+            .scan(last_event_id.map(str::to_owned), |last_event_id, frame| {
                 ready(Some(event_to_item::<T>(last_event_id, frame)))
             })
             .filter_map(ready);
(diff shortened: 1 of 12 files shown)


bmc-http/tests/event_stream_tests.rs
```diff
@@ -272,4 +272,227 @@ mod tests {
...
+            .map(|event| event.last_event_id.as_deref())
+            .collect();
+        assert_eq!(ids, [Some("7"), Some("8")]);
+    }
+
+    #[tokio::test]
+    async fn an_empty_resume_id_is_no_resume_id() {
+        struct WithoutHeader(&'static str);
+
+        impl Match for WithoutHeader {
```
jayzhudev: Seems like this is the same as around line 454?

jayzhudev (dismissed):

bmc-http/src/reqwest.rs
```diff
@@ -1258,20 +1289,80 @@ impl HttpClient for Client {
...
+impl Client {
+    /// Opens the SSE request, `Last-Event-ID` set when resuming, and returns
+    /// its frames: decoded from a body capped at
+    /// [`SseOptions::max_event_bytes`] per event, ended after a fatal error,
+    /// and, when [`SseOptions::idle_timeout`] is set, failed when no frame
+    /// arrives within that window. The window is keyed on frames the server
+    /// sends, so comment heartbeats (which the decoder discards) do NOT reset
+    /// it; a server relying solely on comment heartbeats should not enable
+    /// the idle timeout.
+    // The idle branch matches on `Option<Duration>` where both arms await
```
jayzhudev, line 1342: I just noticed the `//`s here may need to be `///`.

yoks: It looks like AI generated doc comment, and also added regular comment beneath it, first time see something like this. 

Yea this is justification for clippy allow

jayzhudev (approved):

bmc-http/src/reqwest.rs
jayzhudev: Seems legal but I've never thought about this 😀

poroh (approved):

PR #217, opened 2026-09-14
fix(oem-dell): select account handling by iDRAC version

krish-nvidia:
### Summary

This PR makes account behavior immutable for each `AccountService`. Callers pass a configuration to `ServiceRoot::account_service(config)`, and the service exposes only `accounts()`.

Standard systems use `AccountServiceConfig::standard()`. Dell callers select an iDRAC version: iDRAC9 maps internally to fixed slots 3–16, iDRAC10 uses standard `POST`/`DELETE`, and unknown versions return `None`.

### Validation

Hardware testing on three iDRAC9 systems and one iDRAC10 system confirmed that iDRAC9 requires fixed slots while iDRAC10 supports standard `POST` and `DELETE`.


redfish/src/account/item.rs
```diff
@@ -48,12 +47,19 @@ use nv_redfish_core::NavProperty;
 use std::convert::identity;
 use std::sync::Arc;
 
+#[derive(Clone, Copy)]
+pub enum DeletionStrategy {
+    DeleteResource,
+    #[cfg(feature = "oem-dell")]
```
poroh, line 53: Lets avoid oem-dell feature flag here. I see no reasons to have it here and below.

redfish/src/account/mod.rs
```diff
@@ -55,13 +57,68 @@ pub use item::Account;
...
+/// Immutable account behavior selected when loading [`AccountService`].
+#[derive(Clone)]
+pub struct AccountServiceConfig {
+    behavior: AccountServiceBehavior,
+}
+
+#[derive(Clone)]
+enum AccountServiceBehavior {
+    Standard,
+    #[cfg(feature = "oem-dell")]
```
poroh, line 75: The same here. Lets avoid feature-gating here. It just adds more noise.

redfish/src/oem/dell/account.rs
```diff
@@ -0,0 +1,32 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Dell iDRAC account-service behavior.
+
+use crate::account::AccountServiceConfig;
+
+/// Dell iDRAC generation used to select account-management behavior.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum IdracVersion {
```
poroh, line 10: It seems that this PR is incomplete because it doesn't have manager's part of getting IdracVersion.
Also I think that we should follow https://github.com/dell/iDRAC-Redfish-Scripting/blob/4bff644c969ad166b35b20ec3ee640455c9f93e9/Redfish%20Python/CreateIdracUserCsvFileREDFISH.py#L54-L72 and wrap u32 to identify version and use the same logic as there to calculate account config.

poroh (changes requested):

redfish/src/oem/dell/account.rs
krish-nvidia: Yeah you're right, I assumed NICo would fetch the Manager and figure out the iDRAC version before passing the account config but that would put Dell-specific logic in NICo.

Fixed it and NICo now only needs to select the driver and pass the resulting config.

poroh (dismissed):

poroh (approved):

PR #219, opened 2026-09-14
feat(oem-dell): add storage and RAID volume operations

krish-nvidia:
### Summary

This PR adds standard Volume listing and creation, Dell RAID creation, and controller-drive decommissioning.

Dell RAID requests now use the standard generated `VolumeCreate`, `LinksCreate`, and drive reference types. Volume creation handles full entities, Location-only responses, tasks, and empty responses.

The Dell decommission action is defined in `DellStorage_v1.xml` and uses the generated action target, request type, and `ActionAnnotations`. It accepts an optional standard `OperationApplyTime`; omitting it sends Dell’s documented empty payload.

### Validation

Storage mutations were not tested on hardware. The integration tests cover payloads and response handling. Decommission discovery on iDRAC10 remains unverified because the tested system had no eligible controller.


redfish/src/oem/dell/storage_actions.rs
```diff
@@ -0,0 +1,71 @@
...
+
+use serde::{Deserialize, Serialize};
+
+use crate::core::{Action, Bmc, ModificationResponse};
+use crate::schema::settings::OperationApplyTime;
+use crate::schema::storage::OemActions as StorageOemActions;
+use crate::{Error, NvBmc};
+
+#[derive(Debug, Deserialize)]
+struct DellStorageOemActions {
```
poroh, line 16: Should be defined by schema. You can use reference code for NVIDIA OEM actions:

https://github.com/dsx-ai-factory/nv-redfish/blob/main/redfish/src/oem/nvidia/chassis_actions.rs
https://github.com/dsx-ai-factory/nv-redfish/blob/main/schema/oem/nvidia/NvidiaChassis_v1.xml#L74

redfish/src/oem/dell/volumes.rs
```diff
@@ -0,0 +1,93 @@
...
+
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "PascalCase")]
+struct VolumeLinks {
+    drives: Vec<DriveReference>,
+}
+
+/// Typed Dell request for creating a RAID volume.
+#[derive(Debug, Serialize)]
+pub struct DellVolumeCreate {
```
poroh, line 27: This also should be schema-driven.

poroh (changes requested):

redfish/src/oem/dell/volumes.rs
krish-nvidia: For `DellVolumeCreate`, I could not find an existing schema-driven path smilar to the Nvidia action example.

Dell creates a Volume by POSTing to the standard `VolumeCollection`, but its payload requires:

```json
{
  "Name": "...",
  "RAIDType": "RAID1",
  "Links": {
    "Drives": [{"@odata.id": "..."}]
  }
}
```

The generated standard `VolumeCreate` currently omits `Name`, `RAIDType`, and `Links.Drives` because those are read-only/navigation properties in the standard csdl. Dell’s published `DellVolume` schema appears to describe OEM response data, not this create payload.

Should we add Dell-specific csdl annotations plus compiler support for `RequiredOnCreate` navigation properties, or is there an existing schema/compiler pattern I can use?

redfish/src/oem/dell/storage_actions.rs
krish-nvidia: I can move the Dell action container and advertised target into `DellStorage_v1.xml` using the NVIDIA pattern.

The remaining issue is that csdl parameter identifiers cannot contain `@`, and the current compiler would serialize an `OperationApplyTime` parameter as `"OperationApplyTime"` rather than `"@Redfish.OperationApplyTime"`.

Would you prefer that I keep this request payload handwritten while using the generated action target, or add compiler support for redfish operation annotations first?

redfish/src/oem/dell/volumes.rs
poroh: These fields should be added to generation by #222

redfish/src/oem/dell/storage_actions.rs
poroh: This one is covered by #221

poroh (approved):

PR #231, opened 2026-09-19
fix(accounts)!: skip unreadable fixed account slots

jayzhudev:
Fix a regression where `create_account()` aborts when a fixed account slot is missing or returns a malformed response instead of trying a later available slot.

Fixed-slot scans skip HTTP 404/410 and response-parsing failures. Transport failures, HTTP 500 and unclassified errors abort provisioning because they do not establish that the slot is unavailable.

Added implementation-neutral BMC error classification and distinguishes incomplete HTTP bodies from malformed JSON.

### Breaking change

Custom `Bmc` and `HttpClient` implementations must implement `BmcError` for their associated error type:

```rust
impl BmcError for MyError {}
```

The default implementation classifies errors as `BmcErrorClass::Other`, so existing behavior can be preserved without additional logic.


redfish/src/account/collection.rs
```diff
@@ -170,12 +170,17 @@ impl<B: Bmc> AccountCollection<B> {
...
-            // order within the configured bounds.
+            // Collection order is not a stable allocation policy. Inspect each
+            // readable member, then consider disabled numeric slots in ascending
+            // order within the configured bounds. An unreadable slot cannot be
+            // reused safely, but it must not hide later usable slots.
             let mut candidates = Vec::new();
+
             for nav in &self.collection.members {
-                let account = Account::new(&self.bmc, nav, &self.config.account).await?;
+                let Ok(account) = Account::new(&self.bmc, nav, &self.config.account).await else {
```
poroh: Can we skip only specific errors here (4xx & response parse errors)?

jayzhudev: I think that's a good idea. But the error is:

```
Error<B> -> Error::Bmc(e) -> e
```

`e` is `Bmc`'s impl's associated type which has bounds `std::error::Error + Send + Sync`. So checking HTTP codes would require tying this to BMC implementation.

Some of parsing error is in `Error::Json(error)` but BMC parsing error is opaque.

poroh (approved):

